PR #105, opened 2026-03-27
[VW-216] use one-time-tokens for webhook responses instead of api keys

0xcad:
previously a user had to configure an api key for the integrated service on VIPER, and configure a VIPER api key on the integrated service. so for example, a user had to put a Blueflow API key in VIPER and a VIPER API key in Blueflow.

Now, we use one-time-tokens to create unique urls. We pass these to external services, who get authenticated via the unique url, instead of from an API key

## Summary by CodeRabbit

* **New Features**
  * Introduced token-based authentication for integration uploads, replacing API keys with secure, expiring tokens.
  * Added automatic cleanup of expired authentication tokens.
  * Made integration sync frequency configurable via new constant.

* **Removed Features**
  * Removed API key rotation UI and functionality from integrations.

### DIFF
--- a/prisma/migrations/20260326153511_user_tokens/migration.sql
+++ b/prisma/migrations/20260326153511_user_tokens/migration.sql
@@ -1,0 +1,38 @@
+/*
+  Warnings:
+
+  - You are about to drop the column `apiKeyId` on the `integration` table. All the data in the column will be lost.
+
+*/
+-- DropForeignKey
+ALTER TABLE "public"."integration" DROP CONSTRAINT "integration_apiKeyId_fkey";
+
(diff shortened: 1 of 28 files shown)


prisma/migrations/20260326153511_user_tokens/migration.sql
```diff
@@ -0,0 +1,38 @@
+/*
+  Warnings:
+
+  - You are about to drop the column `apiKeyId` on the `integration` table. All the data in the column will be lost.
```
0xcad: drop it >:)

trummelhadron (approved):
Well done!

PR #112, opened 2026-04-12
[VW-218] deviceArtifact hosting in S3

timrcm:
- Related to VW-218, but sort of also a VW-0 cleanup. 
- S3 hosting was originally only setup for remediations in VW-61; this PR extends that to deviceArtifacts for Vanderbilt's use.

## Summary by CodeRabbit

* **New Features**
  * Device artifact creation now returns upload instructions (URLs and required headers) alongside the created artifact.
  * Artifact metadata updates now accept download URLs, content hashes, and file sizes.

* **Documentation**
  * API create endpoint docs updated to include artifact hosting/upload guidance.

* **Tests**
  * Integration tests updated to validate the new response shape and upload instructions.

### DIFF
--- a/src/features/device-artifacts/types.ts
+++ b/src/features/device-artifacts/types.ts
@@ -53,6 +53,17 @@ export type DeviceArtifactResponse = z.infer<
   typeof deviceArtifactResponseSchema
 >;
 
+const uploadInstructionsSchema = z.object({
+  artifactName: z.string(),
+  uploadUrl: z.string().url(),
+  requiredHeader: z.string(),
+});
+
+export const deviceArtifactUploadResponseSchema = z.object({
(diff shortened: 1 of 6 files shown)


src/features/device-artifacts/types.ts
```diff
@@ -53,6 +53,17 @@ export type DeviceArtifactResponse = z.infer<
   typeof deviceArtifactResponseSchema
 >;
 
+const uploadInstructionsSchema = z.object({
+  artifactName: z.string(),
+  uploadUrl: z.string().url(),
+  requiredHeader: z.string(),
+});
```
0xcad, line 60: currently this is duplicated in both device-artifacts and artifacts when they're the same, but I also think this layout makes sense from a code locality perspective. lgtm

0xcad (approved):

PR #113, opened 2026-04-16
[VW-108] Made CPEs and SARIF fields optional for vulnerabilty update API

trummelhadron:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved vulnerability update functionality to support optional CPE specifications and allow partial field updates without requiring all fields to be specified.

### DIFF
--- a/src/features/vulnerabilities/types.ts
+++ b/src/features/vulnerabilities/types.ts
@@ -43,6 +43,22 @@ export const vulnerabilityArrayInputSchema = z.object({
   vulnerabilities: z.array(vulnerabilityInputSchema).nonempty(),
 });
 
+export const vulnerabilityUpdateInputSchema = z.object({
+  cpes: z.array(cpeSchema).min(1).optional(),
+  sarif: z.any().optional(), // JSON data - Prisma JsonValue type
+  cveId: z.string().min(1).nullish(),
+  description: z.string().min(1).nullish(),
+  narrative: z.string().min(1).nullish(),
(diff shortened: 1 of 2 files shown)


src/features/vulnerabilities/types.ts
```diff
@@ -43,6 +43,22 @@ export const vulnerabilityArrayInputSchema = z.object({
...
+  narrative: z.string().min(1).nullish(),
+  impact: z.string().min(1).nullish(),
+  severity: severitySchema.optional(),
+  cvssScore: z.number().min(0).max(10).nullish(),
+  cvssVector: z.string().min(1).nullish(),
+  affectedComponents: z.array(z.string().min(1)).optional(),
+  exploitUri: safeUrlSchema.nullish(),
+  upstreamApi: safeUrlSchema.nullish(),
+  deviceArtifactId: z.string().min(1).nullish(),
+});
```
0xcad:
```
export const vulnerabilityUpdateInputSchema = vulnerabilityInputSchema.extend({
  cpes: vulnerabilityInputSchema.shape.cpes.optional(),
  sarif: vulnerabilityInputSchema.shape.sarif.optional(),
});
```

^This may be cleaner just since then we don't have to keep two copies around (and it maybe more clearly communicates that it's the same as the input schema, we just made two fields optional)

I'll approve regardless though, lmk if you think keeping separate fields is a better approach

0xcad (approved):

PR #132, opened 2026-05-18
[VW-267] added device utilization field to assets db/UI

0xcad:
* Adds device utilization JSON data to asset db and asset types
* Visualizes device utilization information in asset detail UI

This will/could be used by VIPER's AI to schedule remediations, and is something we're exploring with Virta Labs

image

## Summary by CodeRabbit

* **New Features**
  * Device Utilization section is now available in asset details, providing a visual grid display with interactive tooltips to help you monitor and track hourly and daily asset performance metrics.
  * Utilization tracking data is fully integrated throughout all asset management operations, including asset creation, bulk imports, individual updates, and integration synchronization workflows.

[Review Change Stack](https://app.coderabbit.ai/change-stack/PATCH-UPGRADE/viper/pull/132?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/features/assets/components/asset-drawer.tsx
+++ b/src/features/assets/components/asset-drawer.tsx
@@ -73,6 +73,19 @@ function getUtilizationColor(value: number): string {
   return `hsl(120, ${s}%, ${l}%)`;
 }
 
+/*
+ Example utilization data:
+    [
+      {"9": 1, "10": 12, "11": 8, "14": 3, "15": 5},  # 0 = Monday
+      {"9": 1, "10": 8, "11": 15},                    # 1 = Tuesday
+      {"9": 2, "14": 5},                              # 2 = Wednesday
+      {"10": 6, "11": 9, "13": 4},                    # 3 = Thursday
+      {"9": 1, "13": 2},                              # 4 = Friday
+      {},                                             # 5 = Saturday
+      {},                                             # 6 = Sunday
+    ]
+ Tuesday 9am to 10am, the device had 1% utilization throughout the hour, 10am-11am it was 8%, and 11am to 12pm it was 15%, and all other hours of the day it was offline (0% utilization)
+ */
 function AssetUtilizationGrid({
   utilization,
 }: {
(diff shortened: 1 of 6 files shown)


src/features/assets/types.ts
```diff
@@ -61,6 +66,7 @@ export const assetResponseSchema = z.object({
   serialNumber: z.string().nullable(),
   location: z.unknown().nullable(),
   status: assetStatusSchema.nullable(),
+  utilization: z.unknown().nullable(),
```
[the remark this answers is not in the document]
0xcad: prisma returns JsonValue, and TS can't verify what kind of JSON this is. We use unknown here just like what we do for location, even though we have a schema for that, too

src/features/assets/components/asset-drawer.tsx
```diff
@@ -408,6 +539,7 @@ export function AssetDashboardDrawer({
       value: "details",
       label: "Details",
       icon: FileText,
+      rawContent: true,
```
0xcad, line 542: necessary to get horizontal scroll working on overflow

src/features/assets/components/asset-drawer.tsx
```diff
@@ -41,6 +50,110 @@ interface AssetDashboardDrawerProps {
...
+              </td>
+              {data.map((dayData, dayIndex) => {
+                const value = dayData[String(hour)] ?? 0;
+                return (
+                  <td key={dayIndex} className="p-0">
+                    <Tooltip>
+                      <TooltipTrigger asChild>
+                        <button
+                          type="button"
+                          aria-label={`${DAY_NAMES[dayIndex]} ${hour}:00 to ${hour + 1}:00, ${value}% utilization`}
```
t-a-y-l-o-r, line 135: want me to change the inner value to a float? or are we okay with it staying an int% (which reads like a truncated float * 100)

t-a-y-l-o-r (approved):
🚀

src/features/assets/components/asset-drawer.tsx
0xcad: At this point let's keep an int -- float precision for a percentage is questionable, and I agree int is essentially a two-digit Decimal field, which I think is enough precision for this. Int can also be reinterpreted later (you were talking about having a notice count, rather than representing a percentage)

PR #142, opened 2026-06-22
[VW-306] User can forward an email address to the VMP for parsing

0xcad:
User can forward an email address to the VMP via Resend. This kicks off a service that

* Use an LLM to determine if the notification is relevant
* Saves the email, and any attachments, in the database. Uses minio locally for S3 instead of AWS
* Use an LLM to create a normalized Notification in the database
* Use an LLM to extract device groups from the email
* Use an LLM to, from those extracted device groups, either relate those to the Notification, update them with new info, or create them

End result: Notification object on VIPER with relevant DeviceGroup's linked

## Summary by CodeRabbit

* **New Features**
  * Added a comprehensive notifications system with database-backed notification types, reads, and entity mappings.
  * Added Resend email webhook ingestion to create and update notifications from incoming emails.
  * Introduced AI inbox triage to determine relevance, classify notifications, extract affected device groups, and intelligently link/create/update matches.
  * Added support for downloading PDF attachments and storing them in S3.
* **Chores**
  * Updated environment and development setup to support local MinIO via an S3 endpoint.
  * Added required email/processing dependencies and Prisma migration for the notifications schema.


prisma/schema.prisma
```diff
@@ -813,3 +818,157 @@ model Memory {
...
+  channel      NotificationChannel
+  externalId   String? // for an email, this is the resend email id
+  referenceUrl String?
+  raw          Json
+  markdown     String?              @db.Text
+
+  attachments NotificationAttachment[]
+  receivedAt  DateTime               @default(now())
+
+  @@unique([channel, externalId])
```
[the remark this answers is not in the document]
0xcad: IMO, I don't want to add a db guard that exists outside of the prisma schema, if I don't have to, because that'll be really hard for a human to see that there is non-null constrain if the channel is Email

I also don't want to make this non-nullable because I'm sure we'll have endpoints/pages we want to crawl from that don't have a clear defintion of what `externalId` should be

I think it's ok, but can change my mind, to require have externalId be null and just assume we set `externalId` where emails get processed from the Resend ID. Email deduplication by externalId is nice, but if that doesn't work because the Resend API broke and gave us a null email id (already unlikely) it would not be horrible to just have a duplicate NotificationSource. Potentially we can implement an email body hash for deduplication, I'll add a comment for that

prisma/schema.prisma
```diff
@@ -813,3 +818,157 @@ model Memory {
...
+  notification   Notification @relation(fields: [notificationId], references: [id], onDelete: Cascade)
+  userId         String
+  user           User         @relation(fields: [userId], references: [id], onDelete: Cascade)
+  readAt         DateTime     @default(now())
+
+  @@unique([notificationId, userId])
+  @@map("notification_read")
+}
+
+model NotificationAssetMapping {
```
0xcad, line 927: I made essentially duplicate `Notification*Mapping` models for type-safety / db performance, instead of one polymorphic table. This pattern is repeated with `External*Mapping`

src/app/api/email/route.ts
0xcad, line 1: webhook to receive resend emails
not a trpc endpoint / doesn't use trpc-to-openapi because we need the raw body for the webhook verification

src/features/chat/server/routers.ts
0xcad, line 1: removed these so they'd stop showing up in our openapi spec, which other performers see, and since they're no longer necessary. not really in the scope of this ticket/commit but it's easier imo than a second PR

src/inngest/functions/process-inbox-email.ts
0xcad, line 1: This is the main pipeline through which processed emails come through

src/inngest/functions/process-inbox-email.ts
```diff
@@ -0,0 +1,237 @@
...
+
+    // 4. Convert email HTML body to markdown
+    const markdown = await step.run("html-to-markdown", async () => {
+      if (email.html) return turndown.turndown(email.html);
+      return email.text ?? null;
+    });
+
+    // 5. Persist NotificationSource + NotificationAttachment
+    const sourceId = await step.run("save-source", async () => {
+      // DEV: uncomment to reset duplicate so you can replay the same email webhook
```
0xcad, line 132: When testing on Resend, I replay an email webhook. In production, if we see an email with a duplicate Resend ID we want to just ignore it, since that was the webhook re-firing unnecessarily. While testing though we want to see a new email get processed

src/lib/s3.ts
0xcad, line 1: Added the ability to use local minio as an alternative to AWS S3 for testing

mprocs.yaml
0xcad, line 1: pipe with cat because otherwise docker output crashes mprocs

src/lib/s3.ts
```diff
@@ -10,18 +14,22 @@ type ArtifactInput = z.infer<typeof artifactInputSchema>;
...
-  region: process.env.AWS_REGION,
+  region: process.env.AWS_REGION ?? "us-east-1",
   credentials: {
     accessKeyId: process.env.AWS_ACCESS_KEY_ID!,
     secretAccessKey: process.env.AWS_SECRET_ACCESS_KEY!,
   },
+  ...(process.env.S3_ENDPOINT && {
+    endpoint: process.env.S3_ENDPOINT,
+    forcePathStyle: true,
+  }),
```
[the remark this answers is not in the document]
0xcad: I now enforce having AWS_REGION as an env var

src/features/inbox/types.ts
```diff
@@ -0,0 +1,13 @@
+import { z } from "zod";
+
+// TODO: Get a discriminated union of what fields should actually live on Advisory|Recall|UpdateAvailable (see `details`)
```
0xcad, line 3: Here's what I just found poking around Siemens Healthineers' advisory api endpoint

```
{
	"0": {
		"id": 25,
		"advisoryId": "SSA-016040",
		"title": "Insecure Password Encryption Vulnerability in syngo.plaza VB30E",
		"cvssScore": 5.3,
		"version": "1.0",
		"lastUpdated": "2026-02-10T00:00:00.000+00:00",
		"tag": null,
		"modality": [
			"Syngo"
		],
		"cveIds": " CVE-2024-52334",
		"productsAffected": "syngo.plaza VB30E\nAll versions < VB30E_HF07\naffected by CVE-2024-52334",
		"activationDate": "2026-02-10T00:00:00.000+00:00",
		"active": true,
		"severity": "2",
		"lastEmailsSent": null,
		"enableMail": false
	}
}
```

So useful for us here: `advisoryId`, `cvssScore`, `version`, `lastUpdated`, `tag`

`productsAffected` and `cveIds` would live on `NotificationDeviceGroupMapping` and `NotificationVulnerabilityMapping` respectively

src/features/chat/server/routers.ts
```diff
@@ -11,14 +11,6 @@ import {
 export const chatRouter = createTRPCRouter({
   getManyThreads: protectedProcedure
     .input(fetchThreadsSchema)
-    .meta({
```
perrydev17, line 14: double checking, are these removed intentionally?

src/features/inbox/agent/candidate-search.ts
```diff
@@ -0,0 +1,73 @@
...
+};
+
+// Returns top-K candidates per extracted entity.
+async function searchDeviceGroup(
+  extracted: ExtractedDeviceGroup,
+): Promise<DeviceGroupCandidate[]> {
+  const terms = [
+    extracted.cpe,
+    extracted.manufacturer,
+    extracted.modelName,
```
perrydev17, line 33: how about version? also, do we need serial number? I recall it was mentioned in the RFC doc. Agree cpe, manufacturer and modelName are a good start, probably later phase for serial number if it is needed

src/features/inbox/agent/classify.ts
```diff
@@ -0,0 +1,130 @@
...
+
+import "server-only";
+import { ChatAnthropic } from "@langchain/anthropic";
+import { HumanMessage } from "@langchain/core/messages";
+import { z } from "zod";
+import prisma from "@/lib/db";
+import { notificationPayloadSchema } from "../types";
+import { fetchPdfAttachments } from "../utils";
+
+const MODEL = "claude-haiku-4-5-20251001";
```
perrydev17, line 12: curious on how do we decide which Model to use?

src/features/inbox/agent/classify.ts
```diff
@@ -0,0 +1,130 @@
...
+- UpdateAvailable: Vendor firmware or software update announcements
+- Other: Security-relevant notifications that don't fit the above categories
+
+UPSERT DECISION:
+- Return action "update" only when this email is clearly a follow-up or amendment to one of the provided existing notifications (same CVE IDs, same recall number, same product + version, explicit reference to a prior advisory)
+- Return action "create" for anything new or ambiguous
+- When action is "update", set notificationId to the id of the matching existing notification and set reasonWhy to a concise explanation of why this email was matched to that notification`;
+
+// TODO:
+// > Return action "update" only when this email is clearly a follow-up or amendment to one of the provided existing notifications (same CVE IDs, same recall number, same product + version, explicit reference to a prior advisory)
```
perrydev17, line 55: already added in line 50?

src/features/inbox/agent/match.ts
```diff
@@ -0,0 +1,226 @@
...
+const deviceGroupFieldsSchema = z.object({
+  cpe: z.string().nullish(),
+  manufacturer: z.string().nullish(),
+  modelName: z.string().nullish(),
+  version: z.string().nullish(),
+});
+
+const decisionSchema = z.object({
+  kind: z.enum(["deviceGroup"]), // TODO: add vulnerability, remediation, maybe asset...
+  op: z.enum(["link", "update", "create"]),
```
perrydev17, line 26: how about `skip`?

src/features/inbox/types.ts
```diff
@@ -0,0 +1,13 @@
...
+export const notificationPayloadSchema = z.object({
+  type: z.enum(["Advisory", "Recall", "UpdateAvailable", "Other"]),
+  title: z.string(),
+  summary: z.string(),
+  tlp: z
+    .enum(["WHITE", "GREEN", "AMBER", "RED", "CLEAR", "AMBER_STRICT"])
```
perrydev17, line 9: I look into the `https://www.first.org/tlp/` you shared above, is "WHITE" the default?

perrydev17 (approved):
Most of the comments are for context learning

src/features/chat/server/routers.ts
perrydev17: nvm, saw your comment here: https://github.com/PATCH-UPGRADE/viper/pull/142#discussion_r3461216563

src/features/inbox/agent/candidate-search.ts
```diff
@@ -0,0 +1,73 @@
...
+  // or an embedding. For example, if the identified manufacturer is Draeger Inc, and the db manufactuerer is "Draeger", this contains fails
+  for (const term of terms) {
+    or.push(
+      { cpe: { contains: term, mode: "insensitive" } },
+      { manufacturer: { contains: term, mode: "insensitive" } },
+      { modelName: { contains: term, mode: "insensitive" } },
+    );
+  }
+
+  return prisma.deviceGroup.findMany({
```
perrydev17, line 49: await ?

src/features/inbox/agent/candidate-search.ts
0xcad: This code just looks for similar device groups in Viper's db. We are extracting version info, just not looking up device groups by that because of how the SQL lookup is structured

If we did look up version `1.0.0` for example, the SQL query just does and OR on if the fields contains the found version, so this would return *all* device groups that have a version `1.0.0`, many of which I bet  would be totally irrelevant

The improvement I want to get after this PR is using either fuzzy search or an embedding to find similar device groups.

I think serial number would be a field for assets, not device groups (currently I just started with only device groups, adding assets comes later). If we see enough examples in emails from hospitals we see with serialNumber, we can add a serialRange field to DeviceGroupMatchingObject -- the new device group model Mark is working to merge in, not ready yet

src/features/inbox/agent/classify.ts
0xcad: I just picked the cheapest + fastest anthropic one since none of these tasks seemed particularly complex. anthropic because bugcrowd IT approves it

src/features/inbox/agent/classify.ts
0xcad:
```
- Return action "update" only when this email is clearly a follow-up or amendment to one of the provided existing notifications (same CVE IDs, same recall number, same product + version, explicit reference to a prior advisory)
- Return action "create" for anything new or ambiguous
- When action is "update", set notificationId to the id of the matching existing notification and set reasonWhy to a concise explanation of why this email was matched to that notification`;

// TODO:
// > Return action "update" only when this email is clearly a follow-up or amendment to one of the provided existing notifications (same CVE IDs, same recall number, same product + version, explicit reference to a prior advisory)
// This could almost certainly be made deterministic, or at least improved upon
```

Here's lines 50 to 56. What I mean by my comment is that we should have some deterministic non-LLM function check to see if, for example, there is another notification with the exact same CVE IDs. Or the same recall number, or advisory ID. If we can be confident it's a match we just make sure the LLM updates the existing Notification object *always* rather than give it the option to create a new one

src/features/inbox/agent/match.ts
0xcad: Skip is an option effectively, it doesn't have to make a decision. In testing I've seen the `decision` array be entirely empty so I know it can skip

src/features/inbox/types.ts
0xcad: Not sure there's a default per se. White is in version 1 btw, and is now called "Clear". But I keep both because you still see v1 tlp in a lot of places

src/features/inbox/agent/candidate-search.ts
0xcad: https://stackoverflow.com/a/74374895

Pretty sure it's not necessary/inferred in js. The function is awaited on line 68

src/features/inbox/types.ts
perrydev17: noted, thx

PR #147, opened 2026-07-01
[VW-355] Notification DeviceGroupMatching linking

perrydev17:
Jira issue: https://northeastern-patch.atlassian.net/browse/VW-355 

- Link Notification to DeviceGroupMatching
- adding the ability to match vulnerabilites, remediations, and potentially individual assets (not just DeviceGroup)
- extend the existing contains search mechanism (nameMapping) to cover case like  an extracted vendor could be “Draeger Inc”, when our VMP has “Draeger”. Or “Siemens”, when our VMP has “Siemens Healthineers”. 
- Notification FE updates + seed-notifications 

Screenshot 2026-07-02 at 11 49 57 AM
Screenshot 2026-07-02 at 11 50 46 AM
Screenshot 2026-07-02 at 11 50 36 AM

## Summary by CodeRabbit

* **New Features**
  * Inbox matching now supports additional entity types: vulnerabilities, remediations, and assets—alongside device-group matching.
  * Notifications now extract, match, and display coverage using device-group matching identities, including version-range handling and identifier enrichment.
* **Bug Fixes**
  * Device-group links and related coverage are now consistently computed from the updated matching relationship.
  * Affected assets, asset counts, and coverage totals are recalculated using the resolved matching asset data.
* **Chores**
  * Updated database constraints to ensure unique work-order + device-group-matching mappings.


src/features/inbox/agent/candidate-search.ts
```diff
@@ -3,84 +3,140 @@
...
-import type { ExtractedDeviceGroup, ExtractResult } from "./extract";
+import type { ExtractedDeviceGroup, ExtractedVulnerability, ExtractResult } from "./extract";
+import { normalizeName } from "@/lib/router-utils"
 
 const TOP_K = 5;
 
-export type DeviceGroupCandidate = {
+export type DeviceGroupMatchingCandidate = {
   id: string;
-  cpe: string[];
```
0xcad, line 12: What's your opinion of keeping `cpe` and adding a `udi` field?

The fields `cpe` and `udi` are fields on DeviceGroup, not DeviceGroupMatching. 

However, the goal is to still be able to search for them in emails and create a `DeviceGroupMatching` that links to the DeviceGroup that has that `udi`/`cpe` strings

There may be a couple of ways to do this in SQL/prisma. One could just be making a separate query for devicegroup on `cpe`/`udi` if present, then looking to see if there's a related `DeviceGroupMatching`. Do you have any insight on the architecture for this?

src/features/inbox/agent/candidate-search.ts
```diff
@@ -3,84 +3,140 @@
...
+      }
+    })
+  }
+  if(extracted.modelName) {
+    or.push({
+      deviceGroupMatchings: {
+        some: { product: { OR: productNameOr(extracted.modelName)}}
+      }
+    })
+  }
```
0xcad: a little concerned here, but also we don't have a ton of data to test on

but my concern is that if cveId is not present, we'll link manufacturer and model name. we'll get all vulnerabilities whose vendor/product matches, even if those vulnerabilities aren't anywhere in the original email. I think a better case would be to just lookup only cve-ID, but I could be swayed otherwise

src/features/inbox/agent/match.ts
```diff
@@ -143,83 +172,119 @@ export async function applyDecisions(
...
-      }
-      // update the device group and link it to the notification
-      // TODO: consider, we may want to separate this into 'update' and 'update_and_link' actions
-      else if (decision.op === "update") {
-        if (!decision.targetId || !validIds.has(decision.targetId)) {
-          summary.skipped++;
-          continue;
+        // link the device group to the notification
+        if (decision.op === "link") {
+          if (!decision.targetId || !validIds.has(decision.targetId)) {
```
0xcad: TODO: I would separate `validIds` into an object, and here do `validIds.deviceGroups.has(decision.targetId)`.

That way the model is never able to link a vulnerability id where it was supposed to link a device group id

src/features/inbox/agent/match.ts
```diff
@@ -143,83 +172,119 @@ export async function applyDecisions(
...
+        // update the device group and link it to the notification
+        // TODO: consider, we may want to separate this into 'update' and 'update_and_link' actions
+        else if (decision.op === "update") {
+          if (!decision.targetId || !validIds.has(decision.targetId)) {
+            summary.skipped++;
+            continue;
+          }
+          // Vendor/product/version are part of a device group's identity now and
+          // can't be mutated in place; the only safe enrichment is unioning a new
+          // CPE into the existing group's cpe[] array.
```
0xcad, line 384: Actions I would consider trying to take here:

* Can we update the CPE of a device group? (what you're doing now)
* Can we add a new name to nameAliases for vendor/product?
* Can we update the UDI where it didn't previously exist for a DeviceGroup?

prisma/schema.prisma
```diff
@@ -1251,14 +1252,15 @@ model NotificationDeviceGroupMapping {
...
-  deviceGroup    DeviceGroup     @relation(fields: [deviceGroupId], references: [id], onDelete: Cascade)
+  deviceGroupMatchingId String
+  deviceGroupMatching DeviceGroupMatching @relation(fields:[deviceGroupMatchingId], references: [id], onDelete: Cascade)
   confidence     ConfidenceLevel?
   reasonWhy String?
 
-  @@unique([notificationId, deviceGroupId])
-  @@unique([workOrderTicketId, deviceGroupId], map: "ndg_mapping_workorder_devicegroup_key")
+  @@unique([notificationId, deviceGroupMatchingId])
+  @@unique([workOrderTicketId, deviceGroupMatchingId], map: "ndg_mapping_workorder_devicegroup_key")
```
perrydev17: @markglose-bc

src/features/inbox/agent/candidate-search.ts
```diff
@@ -3,84 +3,279 @@
...
+async function searchDeviceGroupMatching(
   extracted: ExtractedDeviceGroup,
-): Promise<DeviceGroupCandidate[]> {
+): Promise<DeviceGroupMatchingCandidate[]> {
+
+  const matched = new Map<string, DeviceGroupMatchingCandidate>();
+
+  // find the owned DeviceGrop with the identifier first, then surface the DeviceGroupMatching
+  // sharing its identy as match
+  const identifierWhere: Prisma.DeviceGroupWhereInput[] = [];
```
perrydev17, line 86: this is to search with cpe/udi via DeviceGroup , then read the vendorId/productId/versionId and use it to find or create deviceGroupMatching

src/features/inbox/agent/candidate-search.ts
```diff
@@ -3,84 +3,279 @@
...
+        where: identity,
+        select: matchingSelect
+      });
+
+      const matching = existingMatching ?? (await prisma.deviceGroupMatching.create({
+        data: identity,
+        select: matchingSelect
+      }));
+
+      matched.set(matching.id, {
```
perrydev17, line 126: because a single notification can supply both kind of evidence(a UDI and a manufacturer name) in the same extracted entry, if it is the same row, the second.set() will overwrites

src/features/inbox/agent/extract.ts
```diff
@@ -9,44 +9,77 @@ import { fetchPdfAttachments } from "../utils";
...
 });
 
-// TODO: extend this, which has just device groups for now, with vulnerabilities
-// and remediations
+export const extractedVulnerabilitySchema = z.object({
+  cveId: z.string().regex(/^CVE-\d{4}-\d{4,}$/i).nullish(), 
+});
+
+export const extractedRemediationSchema = z.object({
+  linkedCveId: z.string().regex(/^CVE-\d{4}-\d{4,}$/i).nullish(),
```
perrydev17: this is added for searchRemediation can search on its own

src/features/inbox/agent/extract.ts
```diff
@@ -9,44 +9,77 @@ import { fetchPdfAttachments } from "../utils";
...
+  description: z.string().nullish()
+});
+
+export const extractedAssetSchema = z.object({
+  ip: z.string().nullish(),
+  hostname: z.string().nullish()
```
perrydev17: worth to add serialNumber, macAddress?

src/features/inbox/agent/extract.ts
```diff
@@ -9,44 +9,77 @@ import { fetchPdfAttachments } from "../utils";
...
 
-// TODO: extend this, which has just device groups for now, with vulnerabilities
-// and remediations
+export const extractedVulnerabilitySchema = z.object({
+  cveId: z.string().regex(/^CVE-\d{4}-\d{4,}$/i).nullish(), 
+});
+
+export const extractedRemediationSchema = z.object({
+  linkedCveId: z.string().regex(/^CVE-\d{4}-\d{4,}$/i).nullish(),
+  description: z.string().nullish()
```
perrydev17: this is not likely getting any match realistically until fuzzy search/embedding in place

src/features/inbox/agent/match.ts
```diff
@@ -46,53 +48,121 @@ export type MatchSummary = {
...
-- "update": the device group matches an existing candidate, but the notification contains additional/missing identifier info worth saving (e.g. a version the record lacks). Set targetId to that candidate's id and put the new values in fields.
-- "create": none of the candidates match. Put the device group's identifiers in fields. A cpe is required to create — if you cannot supply one, prefer "link" to the closest candidate or omit the decision entirely.
+- "update": the device group matches an existing candidate, but the notification contains additional/missing identifier info worth saving (e.g. a versionRange the record lacks). Set targetId to that candidate's id and put the new values in fields.
+- "create": none of the candidates match. Put the device group's identifiers in fields. Manufacturer is required to create — if you cannot supply one, prefer "link" to the closest candidate or omit the decision entirely.
+
+For each extracted Vulnerability, choose exactly ONE action:
+- "link": the vulnerability clearly matches an existing candidate (same CVE id). Set targetId to that candidate's id.
```
perrydev17, line 86: For Vulnerability, Remediation, Assets, should we also allow update/create?

src/features/inbox/agent/candidate-search.ts
perrydev17: ended up doing the query for cpe/udi

src/features/inbox/agent/candidate-search.ts
perrydev17: makes sense, removed

src/features/inbox/agent/match.ts
perrydev17: thanks, updated

src/features/inbox/agent/match.ts
perrydev17: added

prisma/schema.prisma
```diff
@@ -283,7 +283,7 @@ model DeviceGroup {
   // Vulnerabilities / remediations / advisories / device artifacts connect via
   // DeviceGroupMatching now, not a direct relation. notificationMappings is from
   // main's notifications feature and stays.
-  notificationMappings NotificationDeviceGroupMapping[]
+  // notificationMappings NotificationDeviceGroupMapping[]
```
0xcad: You can just delete rather than comment out

Also I know git blame this wasn't from your code, but you can remove the comment on 283-286, I wish I had caught that from Mark's PR (it's just an obvious LLM-ism that references old code where I don't think we should)

src/features/inbox/agent/extract.ts
```diff
@@ -9,44 +9,85 @@ import { fetchPdfAttachments } from "../utils";
...
 const MODEL = "claude-haiku-4-5-20251001";
 
 const SYSTEM_PROMPT = `You are an extraction agent for a hospital cybersecurity platform that reads security notifications (advisories, recalls, update notices) and their PDF attachments.
 
-Your task: extract the DEVICE GROUPS that the notification is about. A device group is a class of affected product, identified by some combination of:
+Your task: extract the following Entities that the notification is about. 
+
+FOR DEVICE GROUPS: A device group is a class of affected product, identified by some combination of:
 - CPE string (e.g. "cpe:2.3:a:vendor:product:1.0:*:*:*:*:*:*:*")
+- UDI: a device-label identifier (distinct from a CPE)
```
0xcad: Worried that this instruction could confuse the model to just pick any identifier it sees

What about
- UDI: The FDA-assigned unique device identifier (e.g, "04048675556176")

^That happens to be the UDI we got in our example email for the ATlan A350: https://accessgudid.nlm.nih.gov/devices/04048675556176

src/features/inbox/agent/match.ts
0xcad: For vulnerability + remediation, yes. For asset, no (good question, that's very ambiguous behavior)

My reasoning is that an asset is something that the hospital owns, and a hospital shouldn't learn of new inventory from their emails.

Vulnerabilities + remediations can be learned from emails/integrations however

src/features/inbox/components/notification-affected-assets-tab.tsx
```diff
@@ -11,17 +11,15 @@ import {
...
-import { deviceGroupLabel, parseLocation } from "@/lib/string-utils";
+import { deviceGroupMatchingLabel, parseLocation } from "@/lib/string-utils";
 import type { NotificationDetailWithRelations } from "../types";
 
 export function NotificationAffectedAssetsTab({
-  deviceGroups,
+  deviceGroupsMatchings,
 }: {
-  deviceGroups: NotificationDetailWithRelations["deviceGroups"];
+  deviceGroupsMatchings: NotificationDetailWithRelations["deviceGroupsMatchings"];
```
0xcad, line 20: All of your type errors are downstream of this line: Property 'deviceGroupsMatchings' does not exist on type 'NotificationDetailWithRelations'.

Seems related to this coderabbit issue: https://github.com/PATCH-UPGRADE/viper/pull/147/changes#r3514907510

Did `npx tsc` work locally? If so, then there's also probably a migrations problem, but i think this is your issue

prisma/schema.prisma
```diff
@@ -1163,8 +1164,8 @@ model Notification {
 
   sources         NotificationSource[]
   reads           NotificationRead[]
-  assets          NotificationAssetMapping[]
-  deviceGroups    NotificationDeviceGroupMapping[]
+  assets          NotificationAssetMapping[] 
+  deviceGroupsMatchings NotificationDeviceGroupMapping[]
```
0xcad, line 1161: Consider renaming to deviceGroupMatchings, here (not deviceGroup_**s**_Matchings)

And see the relevant coderabbit error here: https://github.com/PATCH-UPGRADE/viper/pull/147/changes#r3514907510

src/lib/router-utils.ts
0xcad, line 1: Consider moving fns like `addVendorAlias`, `addProductAlias`, `enrichDeviceGroupIdentifiers`, to a new lib file (one idea: split up agent/match.ts into agent/match/index.ts and agent/match/utils.ts, but feel free to pick something else, too)

My original intention with router-utils was helper functions for trpc endpoints, that are used in multiple `router.ts` files

0xcad (approved):

src/features/inbox/components/notification-affected-assets-tab.tsx
perrydev17: yes it does. I had a typo in one of the types, which caused an error. The other one was the ordering of the imports. I will make sure to run linting before creating a pr next time

prisma/schema.prisma
perrydev17: thanks, I missed this, should have removed instead of comment out

src/features/inbox/agent/match.ts
perrydev17: I added the update for Vulnerability/Remediation. But for create, vulnerability has required fields `userId` and `sarif`, while remediation has `userId`

src/lib/router-utils.ts
perrydev17: moved them to `src/features/inbox/utils.ts`

src/features/inbox/agent/extract.ts
perrydev17: updated

prisma/schema.prisma
perrydev17: updated

PR #155, opened 2026-07-09
[VW-351] Allow User to mark incorrect LLM Notification <-> Device Grouping linking from UI

perrydev17:
ticket: https://northeastern-patch.atlassian.net/browse/VW-351 

## Summary
- Add a mark as incorrect action on the notification's Affected Assets tab. Users can unlink an LLM-matched device group from a notification, with an optional comment explaining why.
- New MatchFeedback table and Rejected confidence value. This is using TargetType approach so it can later extends to Vulnerability/Remediation without schema changes
- Guards the inbox pipeline so a reprocessed notification can never silently relinked

https://github.com/user-attachments/assets/e41c260f-a407-4443-9e78-2564602080ca

## Summary by CodeRabbit

* **New Features**
  * Added a feedback flow for match results, including the ability to mark an item as incorrect, add an optional comment, and save the response.
  * Introduced support for rejected match states so previously dismissed matches are no longer shown as active.
  * Added a new action in the affected assets view to unlink a match with confirmation.

* **Bug Fixes**
  * Prevented rejected device-group matches from being recreated or re-linked.
  * Improved notification views to hide rejected match entries.

### DIFF
--- a/prisma/migrations/20260708193227_match_feedback/migration.sql
+++ b/prisma/migrations/20260708193227_match_feedback/migration.sql
@@ -1,0 +1,34 @@
+-- CreateEnum
+CREATE TYPE "MatchFeedbackTargetType" AS ENUM ('NotificationDeviceGroupMapping', 'NotificationAssetMapping', 'NotificationRemediationMapping', 'NotificationVulnerabilityMapping');
+
+-- CreateEnum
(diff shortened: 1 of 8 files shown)


timrcm (approved):
Looks solid!

PR #158, opened 2026-07-13
[VW-363] Teamplay Fleet Auth

perrydev17:
ticket: https://northeastern-patch.atlassian.net/browse/VW-363

Add authenticated access to Templay Fleet's session-cookie-gated REST API: a Playwright driven login captures a session cookie, cached in a new table and reused across different api calls, with automatic reauth only when a request comes back with 401 or 403.

## Summary by CodeRabbit

* **New Features**
  * Added automated TeamPlay Fleet sign-in and session cookie capture.
  * Added cached integration sessions that automatically reauthenticate when sessions expire or become unauthorized.
  * Added browser automation support for Fleet authentication in hosted and local environments.
  * Added persistent storage for integration sessions, including expiration tracking.
  * Added a diagnostic script for validating Fleet authentication and session behavior.


prisma/schema.prisma
```diff
@@ -1439,4 +1439,16 @@ model MatchFeedback {
   @@index([targetType, targetId])
   @@index([notificationId])
   @@map("match_feedback")
+}
+
+model VendorSession {
```
perrydev17: Is this name potentially confusing?We already use Vendor to refer to a manufacturer, but in this case it would represent a third-party integration provider.

scripts/test-fleet.ts
```diff
@@ -0,0 +1,62 @@
+import {
```
perrydev17, line 2: this is for local testing only, run `npx tsx scripts/test-fleet.ts` should see, some logs including cookies, as well as db record

prisma/schema.prisma
0xcad: consider IntegrationSession?

src/config/teamplay-fleet-auth.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export function getVendorAuthConfig(url: string): VendorAuthConfig | undefined {
+  let host = "";
+  try {
+    host = new URL(url).hostname.toLowerCase();
+  } catch {
+    return undefined;
+  }
+  return VENDOR_AUTH_CONFIGS[host];
+}
```
0xcad: `getVendorAuthConfig`, and the way you have organized this code, implies it's not specific to just teamplay fleet. However, you also call this file `teamplay-fleet-auth.ts` which seems to contradict that

src/features/advisories/server/fleet-advisory-sync.ts
```diff
@@ -0,0 +1,58 @@
+import { getVendorAuthConfig } from "@/config/teamplay-fleet-auth";
+import prisma from "@/lib/db";
+
+export async function fetchWithVendorSession(url: string): Promise<Response> {
+  const config = getVendorAuthConfig(url);
```
0xcad: Again, `fetchWithVendorSession` and `getVendorAuthConfig` implies this can be used for multiple vendors, but the file is called `fleet-advisory-sync.ts`

src/features/advisories/server/fleet-advisory-sync.ts
```diff
@@ -0,0 +1,58 @@
...
+    update: {
+      header: session.header,
+      value: session.value,
+      expiresAt: session.expiresAt,
+    },
+  });
+
+  return fetch(url, {
+    headers: { [session.header]: session.value },
+    signal: AbortSignal.timeout(30000),
```
0xcad: I'd make 30_000 a constant at the top of this file, `const REAUTH_TIMEOUT = 30_000`

src/features/advisories/server/fleet-advisory-sync.ts
0xcad, line 1: I need to create a ticket for this, or just quickly do it myself, but `src/features/advisories` is old code that should be removed

I would move into something like `src/features/integrations/teamplay-fleet/fetch.ts`
* Furthermore, I'd move `src/lib/teamplay-fleet-session-capture.ts` into `src/features/integrations/teamplay-fleet/capture.ts` (or something like that)
* I'd move `src/config/teamplay-fleet-auth.ts` into something like `src/features/integrations/teamplay-fleet/config.ts`

^We should reorganize these files into specific directories so that managing, say, 10 bespoke integrations in the future doesn't become unwieldy. 

Why is this file called `fleet-advisory-sync.ts` if it doesn't have anything to do with advisories? It's just helper functions, is that correct?


I'm less committed to this idea, but consider something like an object-based approach to this. That way we can do something like `TEAMPLAY_INTEGRATION.fetchWithVendorSession()`, `BAXTER_INTEGRATION.fetchWithVendorSession()`, etc, with multiple integrations
* so `fetchWithVendorSession` becomes an instance method, the src/config/teamplay-fleet-auth.ts data would get moved into the constructor


Bottom line:
* Need a better way to organize code among multiple vendors
* Functions/code here implies it's vendor-neutral, despite the fact that it's in a file that references the specific `fleet` vendor

src/lib/teamplay-fleet-session-capture.ts
```diff
@@ -0,0 +1,104 @@
...
+  welcomeUrl: "https://fleet.siemens-healthineers.com/welcome",
+  cookieBannerAcceptSelector:
+    "#CybotCookiebotDialogBodyLevelButtonLevelOptinDeclineAll",
+  welcomeLoginButtonSelector: '[data-cy="btn-login"]',
+  userNameSelector: "#email",
+  continueSelector: "#next_link_container",
+  passwordSelector: "#password",
+  submitSelector: "#btn-login",
+  cookieOrigin: "https://fleet.siemens-healthineers.com",
+  authUrl:
```
0xcad: Is this just a url to test authentication with? If so, I'd add a comment explaining that

0xcad (changes requested):
<deleted>

prisma/schema.prisma
perrydev17: sounds good, updated [here](https://github.com/PATCH-UPGRADE/viper/pull/158/commits/49082323ad4819075c5c0b0ca300deadd9a56f1f)

src/features/advisories/server/fleet-advisory-sync.ts
perrydev17: thanks, yes, I originally went with a more generic approach so we could potentially reuse it for other third-party integration in the future. At the same time, I was a bit hesitant because this isn't an official integration mechanism (e.g API keys or JWT tokens). The basic auth username/password approach may not translate well into other integrations.

As for why this file called `fleet-advisory-sync.ts`, I named it while I was testing against the advisory endpoint (which will eventually map to our Notification flow), but I didn't rename it.

changes: I created an `IntegratoinSessionClient` class that encapsulates the cache lookup and reauth logic. The `features/templay_fleet` folder now only contains Teamplay Fleet specific config and handling.

prisma/migrations/20260714170942_integration_session/migration.sql
0xcad, line 1: I would squash all of your migrations into one

Easiest way I've found to do this with prisma is to simply delete (in your case) the last three migrations, do `npx prisma reset`, `npx prisma migrate dev`, and then you should only have one migration file. Makes future `npx prisma reset` runs faster and reduces unnecessary migrations, since we haven't merged into main yet

0xcad (approved):
Approved, although I would still squash migrations

Very easy to follow, looks clean thank you!

src/features/integrations/integration-session-client.ts
```diff
@@ -0,0 +1,82 @@
+import prisma from "@/lib/db";
+import {
+  grabSessionCookie,
+  SessionLoginConfig,
+} from "./teamplay-fleet/capture";
```
0xcad, line 5: I think again we'll have to do some future re-organizing work adding new vendor platforms but this new structure makes those changes a lot easier to do. We can avoid crossing that bridge until we get another platform we need to authenticate with for integrations (potentially a service account with MedISAO would be next?)

0xcad (approved):

prisma/migrations/20260714170942_integration_session/migration.sql
perrydev17: thanks, did the squash migrations, it is cleaner now

src/features/integrations/integration-session-client.ts
perrydev17: yes, there are still some re-organizing works when adding new vendor, also maybe a few different reauth handling etc

PR #162, opened 2026-07-14
[VW-0] Remove advisories dead code

0xcad:
We previously had an `advisories` feature that was an early prototype of what would become our notifications view

It would:
* Get advisories using the CSAF schema and integrations
* Link advisories to VMP assets

These features are more generally reproduced by the notifications/inbox view. This commit removes that old code to reduce confusion.

## Summary by CodeRabbit

* **Removed Features**
  * Removed the Advisories section, including list and detail pages, sidebar navigation, and related tracking links.
  * Removed support for creating or selecting CSAF integrations.
  * Advisory records and integrations are no longer stored or surfaced in the application.

* **New Features**
  * Added a reusable Traffic Light Protocol badge for displaying classification labels.

* **Updates**
  * Tracking views and linked-item counts now focus on assets, vulnerabilities, and remediations.


src/components/tlp-badge.tsx
0xcad, line 1: TLP still used by notifications

src/features/integrations/components/integrations.tsx
0xcad, line 1: For reference, [CSAF](https://github.com/cisagov/CSAF) is a standard gaining traction to create machine-readable advisories. It's currently used by CISA ICS. Could eventually use it as a reliable way to receive notifications (instead of via email, or AI processing...)

perrydev17 (approved):

PR #151, opened 2026-07-07
[VW-344] Parsing Work Orders From Emails

markglose-bc:
Adding support to parse Work Orders from forwarded Emails

Email Classification pipeline now attempts to classify a `Notification` or a `Work Order` 

Added `Original Email` Description Tab for Work Orders Derived from Emails
image

## Summary by CodeRabbit

* **New Features**
  * Renamed “Tracking” to “Work Orders” across the interface, and added an “Advisories” entry to the sidebar.
  * Improved inbox triage: actionable emails (including PDFs) are converted into structured work orders with scheduling/category/department details and entity linking.
  * Work-order ticket details now include an “Original Email” tab when available.
* **Bug Fixes**
  * Improved matching/linking to correctly handle both work orders and notifications with owner-aware, idempotent linking behavior.
* **Tests**
  * Added coverage for decision application and skipped-link scenarios.

PR #164, opened 2026-07-16
[VW-316] Structured, context-rich hospital impact for triage

buggyhuy:
[VW-316](https://northeastern-patch.atlassian.net/browse/VW-316)

## What changed
- `Notification.hospitalImpact`: `String? @db.Text` → `Json @default("{}")` (structured `{ byline, impactStatement, careAreas, likelihood }`); migration discards old prose. `hospitalImpactSchema` (types.ts) is the source of truth.
- New `triage-context.ts` `gatherTriageContext` feeds triage the resolved hospital reality (linked vulns + VEX statuses, remediations, device groups → assets, clinical workflows, utilization, notes); unions notification- and vuln-level matchings; never returns null (degrades gracefully).
- Relocated `generateWorkflowsMarkdown` to `src/lib/markdown/workflow.ts` + new asset-filtered `workflowClinicalSummary`.
- Redesigned the Hospital Impact card (collapsible byline/impact/care-areas/likelihood) with a `{}`-hides-card guard.
- Restyled shared badges (filled PriorityBadge, Advisory warning icon) per the mock — also affects the inbox list.

## Testing
- 19 tests pass (schema contract + workflow filtering + existing).
- Typecheck clean (one pre-existing unrelated `@sparticuz/chromium` error).
- Playwright-driven UI check of the full + degraded cards against the mock.

## Notes
- Merged current `main` (VW-344 #151) in; `types.ts` conflict resolved, triage auto-adopted VW-344's shared `buildUserMessage`.
- Out of scope (future RFC): Questions/Mitigations agents & tabs, status badge.

[VW-316]: https://northeastern-patch.atlassian.net/browse/VW-316?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Added structured Hospital Impact details (byline, impact statement, care areas, likelihood) with conditional “Hospital Impact” collapsible display.
  * Enhanced notification triage to leverage resolved hospital context and include clinical workflow information.
  * Added an icon to Advisory notification type badges.
* **Style**
  * Updated priority badges with clearer, solid-color styling.
* **Bug Fixes**
  * Prevented Hospital Impact cards from showing when impact data is missing, partial, or empty.
* **Tests**
  * Added validation and Markdown rendering tests for hospital impact and workflow summaries.


prisma/migrations/20260715230256_notification_hospital_impact_json/migration.sql
buggyhuy, line 1: ⚠️ The core, irreversible change — review this first.

**Why:** `hospitalImpact` moves from a free-text string to a structured object, so the column type must go `text` → `jsonb`.

**Why drop + re-add (not an in-place cast):** Postgres/Prisma can't auto-cast arbitrary prose to jsonb. We also *deliberately* chose not to preserve the old strings: wrapping them as `{impactStatement: old}` would leave `byline`/`careAreas`/`likelihood` empty — a worse half-populated state — and triage regenerates the value anyway. So every existing row resets to `{}`.

**What to approve:** this is destructive on prod data (existing impact text is discarded; those notifications show no impact card until re-triaged). Confirm that's acceptable — the data-preserving alternative (a backfill migration) wasn't worth it for values the agent recomputes.

src/features/inbox/types.ts
buggyhuy, line 1: **Why this file matters:** `hospitalImpactSchema` is the *single* definition shared by the triage agent (its structured output) and the UI (parse + guard). One source of truth means the shape can't drift between producer and consumer.

**Why all strings, not enums:** the design mock shows `likelihood`/`careAreas` as descriptive phrases ("Unauthenticated network RCE · PoC exploit code exists"). An enum would throw away that specificity and force awkward buckets. We keep them free-text and instead enforce *grounding* in the prompt (don't invent care areas / numbers) — validation of shape here, truthfulness there.

(The `emailKindSchema` / `workOrderPayloadSchema` below are unrelated — they arrived via the VW-344 merge.)

src/lib/markdown/workflow.ts
buggyhuy, line 1: **Why a shared file exists at all:** triage needs to render "which clinical workflows do the affected assets appear in," but the existing `generateWorkflowsMarkdown` was a *private* helper inside a chat-feature tool. Importing chat internals from the inbox agent would be a wrong-direction cross-feature dependency (and would drag the chat tool's whole dependency graph — network topology, memory rendering — into the inbox path). `src/lib/markdown/` is the neutral home where every other entity→markdown renderer already lives, so workflow renderers belong here.

**Intent of the new `workflowClinicalSummary`:** a *compact, asset-filtered* render — only the workflows whose ASSET nodes reference the affected assets, and just their names + affected steps. The recommendations agent's full-JSON dump is too verbose for Haiku.

**Honest note:** relocating the *existing* `generateWorkflowsMarkdown` here too (vs. leaving it in the chat tool) is optional polish for cohesion — it shares the `serializeWorkflow`/type scaffolding with the new fn. Happy to leave it in place if you'd prefer a smaller diff.

src/features/inbox/components/notification-overview-tab.tsx
buggyhuy, line 1: **Why the `safeParse` guard is the critical line here:** `hospitalImpact` is jsonb defaulting to `{}` (per the ticket). `{}` is *truthy* in JS, so the previous `notification.hospitalImpact && …` guard would render an empty card on every un-triaged notification. We parse with `hospitalImpactSchema` and treat a failed parse / empty object as "not triaged yet" → hide the card entirely.

**Intent of the layout:** match the design mock — a collapsible section, a **bold byline** (the one-line takeaway a hospital admin reads first), the impact statement, then the CARE AREAS / LIKELIHOOD boxes. This card is the whole user-facing payoff of the ticket.

src/components/priority-badge.tsx
buggyhuy, line 1: **Why:** the design mock renders priority as a prominent solid/filled pill, not tinted text, so it reads as a status at a glance.

**Blast radius to review (the reason this is worth a comment):** `PriorityBadge` is a *shared* component — this restyle also changes the inbox **list** (`columns.tsx`) and anywhere else it renders. That's intended (design consistency), but it means the change is bigger than "the detail page."

**Accessibility note:** color pairs keep text legible on the fill (Monitor uses dark text on yellow rather than white), which matters given the Section 508 requirement called out in CLAUDE.md.

src/features/inbox/components/notification-type-badge.tsx
buggyhuy, line 1: **Why:** the mock puts a warning icon on the Advisory badge so the most safety-relevant notification type is scannable at a glance.

**Scope choice:** only Advisory gets the icon (the type most likely to need action); the others are unchanged rather than inventing icons for each without a design reference.

**Blast radius:** shared component — like `PriorityBadge`, this also affects the inbox list, not just the detail header.

scripts/seed-notifications.ts
0xcad, line 1: I had to modify seed-notifications.ts pretty heavily for my PR, you'll have a merge conflict here I believe, but looks good

src/components/priority-badge.tsx
```diff
@@ -4,33 +4,32 @@ import type React from "react";
 import type { Priority } from "@/generated/prisma";
 import { cn } from "@/lib/utils";
 
+// Filled (solid) badge styles so priority reads as a prominent status pill.
+// Darker shades keep white-on-color text legible (Monitor uses dark text on
+// yellow, which reads better than white).
```
0xcad: In general I think most comments like these add noise to the project, I would remove these
We don't need to comment our design motivation (especially just color choice) in a component unless something is unintuitive, or maybe there's a reason to use an exact color ([TLP](https://www.first.org/tlp/) is one example, where the color is according to a spec)

src/features/chat/viper-agent/tools/get-recommendations-context.ts
```diff
@@ -71,29 +71,12 @@ function generateVulnAssetRemMap(
 
 // ─── Workflow markdown ────────────────────────────────────────────────────────
 
+// generateWorkflowsMarkdown now lives in @/lib/markdown/workflow (shared with
+// the inbox triage agent). The type below is still used by generateContextMarkdown.
```
0xcad: same on comment but I assume that's just due to draft PR?

src/features/inbox/agent/triage/context.ts
0xcad, line 1: Just to be consistent with current file organization (see how `vex` is structured) I would use /agent/triage/context.ts

src/features/inbox/agent/triage/context.ts
```diff
@@ -0,0 +1,377 @@
...
+        },
+      },
+      vulnerabilities: {
+        include: {
+          vulnerability: {
+            include: {
+              deviceGroupMatchings: {
+                include: { vendor: true, product: true, version: true },
+              },
+              remediations: true,
```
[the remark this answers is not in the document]
0xcad: @buggyhuy

src/features/inbox/agent/triage/context.ts
```diff
@@ -0,0 +1,377 @@
...
+    assetIds: affectedAssetIds,
+  });
+
+  // Clinical workflows that include the affected assets (compact render).
+  const workflows =
+    affectedAssetIds.length > 0
+      ? await prisma.workflow.findMany({
+          include: { nodes: true, connections: true },
+        })
+      : [];
```
0xcad, line 153: Out of scope for this ticket but we could assign assets directly to nodes themselves in the db. That way we don't have to filter through *all* workflows -- but also |Workflows| is not large so this certainly works until proven otherwise

src/features/inbox/agent/triage/context.ts
```diff
@@ -0,0 +1,377 @@
...
+    );
+  }
+
+  // Care-area grounding: where the affected assets live and what they are.
+  // The triage agent may only phrase `careAreas` from these lines.
+  if (args.affectedAssets.length > 0) {
+    const careLines = [
+      ...new Set(
+        args.affectedAssets.map(({ asset, groupLabel }) => {
+          const loc = parseLocation(asset.location);
```
0xcad: location has room numbers. I'd consider only adding `[loc.facility?, loc.building?, loc.floor?].join('/')`

src/features/inbox/types.ts
```diff
@@ -110,6 +110,22 @@ export const notificationPayloadSchema = z.object({
 
 export type NotificationPayload = z.infer<typeof notificationPayloadSchema>;
 
+// Structured hospital-impact output produced by the triage agent and stored on
+// Notification.hospitalImpact (jsonb). Single source of truth for both the
+// agent's structured output and the UI parse. All fields are plain strings —
+// `likelihood` is a free-text descriptor (e.g. "Unauthenticated network RCE ·
+// PoC exploit code exists"), not an enum; `careAreas` is prompt-grounded from
+// the affected assets' locations/roles/device groups (e.g. "Radiology — MRI,
+// CT, Reading Room").
```
0xcad: For example, I'd consider this unnecessary too

0xcad (changes requested):
surface level comments on comments you added + file structuring mostly

requested change on care areas locations (get rid of room numbers from location to reduce what gets sent to LLM context)

scripts/seed-notifications.ts
buggyhuy: Sounds good, I will take care of the conflicts

src/features/chat/viper-agent/tools/get-recommendations-context.ts
buggyhuy: yes, I am still testing out rules for my agent. This is great feedback loop

src/components/priority-badge.tsx
buggyhuy: I will remove them

src/features/inbox/types.ts
buggyhuy: yes I will remove them.

src/features/inbox/agent/triage/context.ts
buggyhuy, line 1: **Read this first — the heart of the change (new file).**

**Why a new module instead of reusing `vex/context.ts`:** the VEX agent's `gatherVexContext` is issue-shaped — it scopes to *vulnerability-level* matchings and returns `null` when there are no vulns/issues. Triage runs for *every* notification (recalls, update notices, "Other"), so it needs the opposite defaults, which is exactly what this file provides:
- it **unions notification-level AND vuln-level** device-group matchings (a recall can link devices with no CVE), and
- it **never returns null** — with no linked assets it still returns a degraded context so triage always has something to reason over.

**Intent:** hand triage the resolved hospital reality (assets, the clinical workflows they're in, utilization, the VEX verdicts computed one step earlier, notes) so impact/priority reflect real exposure instead of a guess from the email text.

**Design choice:** reuses the shared `src/lib/markdown` renderers rather than duplicating them, and deliberately does *not* refactor the VEX agent (critical path, own tests) — hence some intentional query overlap.

src/features/inbox/agent/triage/index.ts
buggyhuy, line 1: **Why the rewrite:** triage used to emit a free-text blob from the email alone and — despite a comment claiming otherwise — never actually read the VEX verdicts produced one step earlier. It now consumes `gatherTriageContext` and emits the structured `hospitalImpact`.

**The load-bearing part to scrutinize is the prompt's grounding rules:** `careAreas` must be phrased only from the provided context's locations/roles/device-groups (no invented ward names on a clinical, admin-facing card), and `likelihood` must summarize the actual CVSS/EPSS/KEV/VEX evidence (never invent numbers).

**Why maxTokens 1024 → 2048:** the structured object + richer context needs headroom. Kept Haiku since the deterministic context does the heavy lifting.

(File moved to `triage/index.ts` to mirror the `vex/` folder layout; the public import `@/features/inbox/agent/triage` is unchanged.)

src/features/inbox/agent/triage/context.ts
buggyhuy: Sounds good

src/features/inbox/agent/triage/context.ts
buggyhuy: I agree. We can open a separate ticket to address this problem.

src/features/inbox/agent/triage/context.ts
buggyhuy: done

0xcad (approved):

PR #168, opened 2026-07-17
[VW-362] Create Notes on Documentation Upload

markglose-bc:
## Summary by CodeRabbit

* **New Features**
  * Automatically extracts security-relevant notes from uploaded PDF documentation.
  * Processes large documents in sections while avoiding duplicate notes across chunks.
  * Creates scoped notes for matching devices and resolves newly created filters when needed.
  * Triggers note extraction when device artifacts are newly created or synchronized.
* **Bug Fixes**
  * Improved handling of missing or delayed PDF uploads during extraction.
  * Refined available filtering options for assets, vulnerabilities, and remediations.
* **Tests**
  * Expanded coverage for document processing, text chunking, note planning, and schema validation.

### DIFF
--- a/src/features/device-artifacts/server/routers.ts
+++ b/src/features/device-artifacts/server/routers.ts
@@ -1,5 +1,6 @@
 import { z } from "zod";
 import { type Prisma, ResourceType } from "@/generated/prisma";
+import { requestArtifactNoteExtraction } from "@/inngest/functions/extract-artifact-notes";
 import prisma from "@/lib/db";
 import { paginationInputSchema } from "@/lib/pagination";
 import {
@@ -211,6 +212,10 @@ export const deviceArtifactsRouter = createTRPCRouter({
         });
       });
 
+      // Kick off note extraction from any uploaded PDF documentation. The job
+      // polls S3, so it tolerates hosted uploads that finish after this returns.
+      await requestArtifactNoteExtraction(result.id);
+
       return {
         deviceArtifact: transformArtifactWrapper(result),
         uploadInstructions: uploadInstructions,
@@ -267,6 +272,8 @@ export const deviceArtifactsRouter = createTRPCRouter({
               },
             };
           },
+          // Extract notes from any uploaded PDF docs on newly-synced artifacts.
+          onItemCreated: (id) => requestArtifactNoteExtraction(id),
         },
         input,
         userId,
@@ -343,6 +350,12 @@ export const deviceArtifactsRouter = createTRPCRouter({
         data,
         include: deviceArtifactInclude,
       });
+
+      // A CPE change re-scopes the artifact to different device group
+      // matching(s); re-run extraction so notes scope to the new matching(s).
+      if (cpe) {
+        await requestArtifactNoteExtraction(id);
+      }
 
       return transformArtifactWrapper(deviceArtifact);
     }),
--- a/src/features/device-artifacts/server/routers.ts
+++ b/src/features/device-artifacts/server/routers.ts
@@ -351,12 +351,6 @@ export const deviceArtifactsRouter = createTRPCRouter({
         include: deviceArtifactInclude,
       });
 
-      // A CPE change re-scopes the artifact to different device group
-      // matching(s); re-run extraction so notes scope to the new matching(s).
-      if (cpe) {
-        await requestArtifactNoteExtraction(id);
-      }
-
       return transformArtifactWrapper(deviceArtifact);
     }),
 });
--- a/src/features/notes/agent/extract-notes.ts
+++ b/src/features/notes/agent/extract-notes.ts
@@ -1,0 +1,79 @@
+// Reads vendor device documentation (PDF) and emits structured create/update
+// operations against Note records. The PDF rides along as a native Anthropic
+// `document` block (no text extraction / chunking) — see @/lib/agent-messages.
+
+import "server-only";
+import { ChatAnthropic } from "@langchain/anthropic";
+import { z } from "zod";
+import { buildUserMessage, type PdfAttachment } from "@/lib/agent-messages";
+
+const MODEL = "claude-haiku-4-5-20251001";
+
+/** A single note the model wants to create, or an update to an existing one. */
+export const noteOpSchema = z.object({
+  action: z.enum(["create", "update"]),
+  // Present (and required) only for `update`: the id of an existing note from
+  // the provided list. Ignored for `create`.
+  noteId: z.string().nullish(),
+  text: z.string().min(1),
+});
+export type NoteOp = z.infer<typeof noteOpSchema>;
+
+export const noteOpsSchema = z.object({
+  notes: z.array(noteOpSchema),
+});
+export type NoteOps = z.infer<typeof noteOpsSchema>;
+
+/** An existing note offered to the model as an update/dedupe candidate. */
+export type ExistingNote = { id: string; text: string };
+
+const SYSTEM_PROMPT = `You extract durable security facts about a medical device from its vendor documentation.
+
+Focus ONLY on facts useful for vulnerability management and hardening, such as:
+- Device hardening guidance and secure-configuration options
+- Attack surface: exposed network services, ports, protocols, wireless/physical interfaces
+- Authentication: default credentials, password policy, certificate/key handling
+- Network exposure and segmentation guidance
+- Compensating controls, mitigations, and known operational constraints
+- Patch/update mechanism and any constraints (e.g. requires downtime, vendor-signed)
+
+Rules:
+- Emit each fact as its own concise, self-contained note (one atomic fact per note). No preamble.
+- Ignore marketing copy, legal boilerplate, and clinical/usage instructions unrelated to security.
+- You are given a list of EXISTING NOTES already attached to this device. If a fact you find is
+  already represented by one of them, emit an "update" op with that note's id and improved text
+  INSTEAD of creating a near-duplicate. Only "update" an id that appears in the list.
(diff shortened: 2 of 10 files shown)


src/features/notes/agent/extract-notes.ts
markglose-bc, line 1: Main file to extract PDF and convert it to a document block.

@0xcad The Jira Ticket requested that we

```
- Turn a PDF into text

- Group sections of the PDF into slightly-overlapping chunks of X tokens
```

Claude recommended that we leverage the Anthropic `document` block vs the proposed solution. I'm still relatively unfamiliar to the nitty gritty of the AI process, so I'll defer to your knowledge, are we okay with this or do we prefer the solution on the Jira ticket?

src/features/notes/server/artifact-notes.ts
markglose-bc, line 1: Plumbing for processing pdfs, broken out for testability.

src/features/device-artifacts/server/routers.ts
```diff
@@ -211,6 +212,10 @@ export const deviceArtifactsRouter = createTRPCRouter({
         });
       });
 
+      // Kick off note extraction from any uploaded PDF documentation. The job
+      // polls S3, so it tolerates hosted uploads that finish after this returns.
+      await requestArtifactNoteExtraction(result.id);
```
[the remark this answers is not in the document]
markglose-bc: This might be worth breaking out into a chore ticket for all the inngest events

src/lib/entity-filter.ts
```diff
@@ -128,7 +128,6 @@ const deviceGroupWhere = buildWhereSchema({
   versionId: stringFilter,
   versionStatus: stringFilter,
   udi: stringFilter,
-  helmSbomId: stringFilter,
```
markglose-bc, line 131: Cleaning up some nonblocking request from the previous PR

src/features/device-artifacts/server/routers.ts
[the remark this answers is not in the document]
0xcad: I'll create a very low priority ticket but I don't think we need to address this now. As long as the inngest event is triggered we should get retries/logs about failures. If we fail to trigger the inngest event entirely, I agree the job just gets lost (but the API should never throw an error, we just console log)

src/features/device-artifacts/server/routers.ts
```diff
@@ -344,6 +351,12 @@ export const deviceArtifactsRouter = createTRPCRouter({
         include: deviceArtifactInclude,
       });
 
+      // A CPE change re-scopes the artifact to different device group
+      // matching(s); re-run extraction so notes scope to the new matching(s).
+      if (cpe) {
+        await requestArtifactNoteExtraction(id);
+      }
+
```
0xcad: Requested Change: We can delete this and ignore this case for now.

This code would just trigger the inngest event again and create duplicate notes (instead of re-assigning any of the existing notes), is that correct? If so that's not great behavior because it just runs a potentially token-costly operation again without cleaning up from before.

src/features/device-artifacts/server/routers.ts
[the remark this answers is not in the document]
0xcad: Idea: If in the future we create an integration for teamplay Fleet where we download PDF's from their platform, we could create a prisma client extension instead of calling `requestArtifactNoteExtraction` in the API endpoint. That way we'd always have the inngest event running, no matter how a deviceArtifact was created

Would also remove the need to have an `onItemCreated` function in the integration sync code.

This is the pattern we use for vulnerabilities btw -- `vulnerabilityExtension` calls an inngest event whenever vulns are created

src/features/notes/agent/extract-notes.ts
0xcad, line 1: @perrydev17 any thoughts on re-using this for VW-376 in the future?

0xcad: I have an example PDF from teamplay fleet that's 68 pages long. That's going to exceed the token window. So we do have to chunk up the PDF into windows that are processable for a haiku agent. We can't chunk up raw base64 without breaking a PDF, which is why I recommended turning the PDF into text first

I do not believe this will scale to a large PDF, requested change unless you have other info

src/features/notes/agent/extract-notes.ts
```diff
@@ -0,0 +1,79 @@
...
+  }).withStructuredOutput(noteOpsSchema);
+
+  const prompt = [
+    "Extract security-relevant notes from the attached device documentation.",
+    existingNotesBlock(existingNotes),
+  ].join("\n\n");
+
+  return model.invoke([
+    { role: "system", content: SYSTEM_PROMPT },
+    buildUserMessage(prompt, pdfs),
```
0xcad: This is the part where `pdfs` is going to overwhelm the token window of a smaller model

src/features/notes/server/artifact-notes.ts
```diff
@@ -0,0 +1,172 @@
...
+          data: {
+            noteId: note.id,
+            label,
+            targetModel: "DEVICE_GROUP_MATCHING",
+            filter: { id: { in: matchingIds } },
+          },
+          select: { id: true },
+        });
+        createdFilterIds.push(filter.id);
+      }
```
0xcad: Requested change: I would add an if condition here, where if there is only one matchingId (i.e, the artifact is only linked to one DGM), make the notes get attached directly via instanceId instead of EntityFilter. Otherwise we unnecessarily create an EntityFilter that an inngest cron job has to resolve instead

0xcad (changes requested):
3 requested changes, see comments. I'll send you something in slack to test on

src/features/notes/agent/extract-notes.ts
perrydev17: so the flow will be, if a user has a match feedback, create a note, and re-use this extract-notes agent to proceed?

src/features/device-artifacts/server/routers.ts
markglose-bc: deleted!

src/features/notes/agent/extract-notes.ts
markglose-bc: yep yep, tested with the pdf you sent in slack and the chunking model is much more performant thank you!

src/features/notes/server/artifact-notes.ts
markglose-bc: gotcha!

src/features/notes/server/artifact-notes.ts
```diff
@@ -3,10 +3,19 @@
...
 import type { NoteOp } from "@/features/notes/agent/extract-notes";
 import { requestEntityFilterResolve } from "@/inngest/functions/resolve-entity-filters";
-import type { PdfAttachment } from "@/lib/agent-messages";
 import prisma from "@/lib/db";
 import { downloadBufferFromS3, keyFromDownloadUrl } from "@/lib/s3";
 
+// Text-chunk sizing. A vendor manual can run dozens of pages (e.g. the 68-page
+// Siemens Cios manual ~= 23k tokens of text), which as image-based PDF document
+// blocks would swamp Haiku's window; extracting text and chunking keeps each
+// agent call small and cheap. ~40k chars ~= 10k tokens, well within context.
```
0xcad: I would leave out the specific manual we used for testing in our open source repo

src/features/notes/server/artifact-notes.ts
```diff
@@ -30,8 +38,47 @@ function isMissingObjectError(err: unknown): boolean {
   return e?.name === "NoSuchKey" || e?.$metadata?.httpStatusCode === 404;
 }
 
+/** Extract plain text from a PDF buffer. unpdf is imported lazily so unit tests
+ * that only touch the pure helpers don't pull in the pdf.js bundle. */
+export async function pdfBufferToText(buffer: Buffer): Promise<string> {
+  const { extractText, getDocumentProxy } = await import("unpdf");
+  const pdf = await getDocumentProxy(new Uint8Array(buffer));
+  const { text } = await extractText(pdf, { mergePages: true });
```
0xcad, line 45: [Draeger Inc Atlan 350, A350 XL Anesthesia Workstations 2.pdf](https://github.com/user-attachments/files/30231936/Draeger.Inc.Atlan.350.A350.XL.Anesthesia.Workstations.2.pdf)

Does this work on a document that needs OCR (like this one which I'll attach again here, but it's also in the [email processing RFC](https://docs.google.com/document/d/1gpFPre3s6aWn-8rI8iJ0YSunvygcQm6mQk9zL6TCxkY/edit?tab=t.2v50peaz17qs)

We can move that to a new ticket if needed, just want to track if unpdf does this or not already

0xcad (approved):
1. Does this work with non-searchable PDF's? Is an OCR step necessary (future ticket?)
2. Adjust comment to not include specific manual we used in a test case...

src/inngest/functions/extract-artifact-notes.ts
```diff
@@ -0,0 +1,168 @@
...
+          id: true,
+          userId: true,
+          role: true,
+          deviceGroupMatchings: { select: { id: true } },
+        },
+      }),
+    );
+
+    if (!artifact) return { skipped: true, reason: "artifact not found" };
+    const matchingIds = artifact.deviceGroupMatchings.map((m) => m.id);
```
0xcad, line 69: Should probably add a quick gate right after this to ensure that matchingIds is non empty since I don't think that's guaranteed

src/features/notes/server/artifact-notes.ts
markglose-bc: Looks like it does not work on a document that requires OCR

PR #170, opened 2026-07-20
[VW-379] Pass PDF attachments through Inngest steps instead of refetching

buggyhuy:
[VW-379](https://northeastern-patch.atlassian.net/browse/VW-379)

Processing one inbound email with one PDF moved the same bytes across the network **six times** — 2 Resend downloads, 1 S3 PUT, 3 S3 GETs — with no cache anywhere. Inngest's per-step retries multiplied it: each download sits inside a `step.run` co-located with a long, failure-prone LLM call, so one Anthropic 429 in `triage-notification` re-downloaded the PDF.

## What changed
- `classify-email` already downloads every PDF (it judges the attachment, not just the body), so it now **returns those buffers in its step output** instead of discarding them at the step boundary.
- `upload-attachments` reuses that buffer — matched by Resend attachment `id`, not filename, since filenames collide — skipping both the signed-URL mint and the re-download.
- The four agents (`classifyNotification`, `extractEntities`, `extractWorkOrder`, `triageNotification`) take an optional PDF array that replaces their `fetchPdfAttachments(sourceId)` S3 GET.
- `fetchPdfAttachmentsFromResend` now returns `{ pdfs, complete }` and includes the Resend `id`.

Two conditions gate the pass-through; failing either yields `null`, which restores today's flow byte-for-byte at every consumer:
- **Size** — total base64 length ≤ `INLINE_ATTACHMENT_BUDGET` (4,000,000), under Inngest's 4MiB step-output cap.
- **Completeness** — every PDF downloaded. A partial set is refused rather than inlined; silently dropping a PDF the S3 fallback could still have retrieved is a behavior regression disguised as an optimization.

| | Resend downloads | S3 PUT | S3 GET |
|---|---|---|---|
| Before | 2 | 1 | 3 |
| After, under budget | 1 | 1 | **0** |
| After, over budget or partial | 2 | 1 | 3 (unchanged) |

## Testing

Measured how many times a single inbound email's PDF crosses the network.
S3 counts are real MinIO requests (`mc admin trace`), reconciled against the bucket.

### Before → After

One email, one PDF under the budget:

| | Resend downloads | S3 PUT | S3 GET | total |
|---|---:|---:|---:|---:|
| **Before** | 2 | 1 | 3 | **6** |
| **After** | 1 | 1 | **0** | **2** |

Over the budget (base64 > 4,000,000 chars ≈ 2.86 MiB raw) it falls back to the old
path, unchanged by design.

Retries improve too: each agent step used to re-fetch the PDF on every retry, so one
Anthropic 429 cost 3 more S3 reads. Now a retried step costs **0**.

### Runs

| run | source | PDFs | bytes | path | S3 GET |
|---|---|---:|---:|---|---:|
| small | mocked | 1 | 61,440 | inlined | **0** |
| large | mocked | 1 | 3,500,000 | fallback | 3 |
| retry (forced step failure) | mocked | 1 | 90,083 | inlined | **0** |
| `8cf5010a` | **real email** | 4 | 78,068 | inlined | **0** |
| `ffcc9339` | **real email** | 4 | 78,068 | inlined | **0** |
| `6c71839c` | **real email** | 3 | 5,335,072 | fallback | 9 |

## Notes
- The budget measures the **actual** `base64.length` of the downloaded PDFs, not Resend's reported `size`. The check runs after the download, so the exact byte count that gets serialized is already in hand, and base64 is escape-free single-byte in JSON — so `.length` *is* the contribution to the step output. An earlier revision estimated from `size` (undocumented unit) with a 4/3 expansion margin; the second commit drops that, along with an unreachable missing-size guard and a duplicated PDF filter.
- **Trade-off to watch**: the inlined payload re-rides with the accumulated run state on every later step invocation (~7 on the notification path). Within Inngest's 32MiB run-state limit, but it trades 3 S3 GETs for ~7 state transfers. If step latency regresses on a large-but-under-budget PDF, drop the constant to ~1MB — it's the only knob.
- **Pre-existing, flagged not fixed**: `keyFromDownloadUrl` (`src/lib/s3.ts`) prefix-matches against the *current* `S3_ENDPOINT`/`S3_BUCKET_NAME`; on mismatch it returns the whole URL as the key, GetObject 404s, and `fetchPdfAttachments` swallows it — agents then run with no attachment at all. Inlining now masks this on the small path.
- Out of scope: the `upload-attachments` retry orphan (hash dedupe reads a row not written until `save-source`, so a retry re-PUTs under a fresh UUID key) — needs content-addressed keys, separate ticket. Also `resend.emails.receiving.attachments.list`, which would batch the remaining one-at-a-time URL lookups but needs pagination handling.

[VW-379]: https://northeastern-patch.atlassian.net/browse/VW-379?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * PDF attachments can now be processed inline when they fit within the supported size limit.
  * Email classification, triage, entity extraction, and work-order processing can reuse pre-loaded inline PDF data to avoid redundant fetching.
  * Attachment handling falls back to secure, time-limited downloads when inline processing isn’t available.
* **Bug Fixes**
  * Prevents using partially downloaded PDF sets by detecting incomplete PDF availability and avoiding downstream processing when downloads fail.
* **Documentation**
  * Updated onboarding with a detailed “Email Pipeline (Resend) Setup” guide and troubleshooting steps.

### DIFF
--- a/src/features/inbox/utils.ts
+++ b/src/features/inbox/utils.ts
@@ -14,39 +14,18 @@ const isPdf = (a: {
   );
 
 /**
- * Ceiling on the summed Resend-reported `size` of an email's PDFs for them to
- * be carried through Inngest step state instead of re-fetched from S3.
- *
- * Resend does not document the unit, so we assume the worst case (decoded
- * bytes). Base64 inflates ~4/3, and the encoded string is what rides in the
- * step output, which Inngest caps at 4MiB — so 3MB here lands at ~4MB encoded.
+ * Ceiling on the total base64 length of an email's PDFs for them to be carried
+ * through Inngest step state instead of re-fetched from S3. The encoded string
+ * is what rides in the step output, which Inngest caps at 4MiB (4,194,304) —
+ * the remainder is headroom for the rest of the step's JSON.
  */
-export const INLINE_ATTACHMENT_BUDGET = 3_000_000;
-
-type ResendAttachmentMeta = {
-  filename?: string | null;
-  content_type?: string | null;
-  size?: number | null;
-};
+export const INLINE_ATTACHMENT_BUDGET = 4_000_000;
 
 /** A PDF carried through step state, keyed back to its Resend attachment. */
 export type InlinePdfAttachment = PdfAttachment & { id: string };
 
-/**
- * Whether an email's PDFs are small enough to pass through step state. An
- * attachment with no reported size is treated as too big — the S3 fallback is
- * always correct, so unknowns resolve that way.
- */
-export function pdfsFitInlineBudget(
-  attachments: ResendAttachmentMeta[],
-): boolean {
-  const pdfs = attachments.filter((a) =>
-    isPdf({ filename: a.filename, contentType: a.content_type }),
-  );
-  if (pdfs.length === 0) return true;
-  if (pdfs.some((a) => typeof a.size !== "number")) return false;
-
-  const total = pdfs.reduce((sum, a) => sum + (a.size ?? 0), 0);
+export function pdfsFitInlineBudget(pdfs: PdfAttachment[]): boolean {
+  const total = pdfs.reduce((sum, pdf) => sum + pdf.base64.length, 0);
   return total <= INLINE_ATTACHMENT_BUDGET;
 }
 
--- a/src/features/inbox/utils.ts
+++ b/src/features/inbox/utils.ts
@@ -1,4 +1,5 @@
 import "server-only";
+import type { PdfAttachment } from "@/lib/agent-messages";
 import prisma from "@/lib/db";
 import { normalizeName } from "@/lib/router-utils";
 import { downloadBufferFromS3, keyFromDownloadUrl } from "@/lib/s3";
@@ -13,9 +14,50 @@ const isPdf = (a: {
   );
 
 /**
+ * Ceiling on the summed Resend-reported `size` of an email's PDFs for them to
+ * be carried through Inngest step state instead of re-fetched from S3.
+ *
+ * Resend does not document the unit, so we assume the worst case (decoded
+ * bytes). Base64 inflates ~4/3, and the encoded string is what rides in the
+ * step output, which Inngest caps at 4MiB — so 3MB here lands at ~4MB encoded.
+ */
+export const INLINE_ATTACHMENT_BUDGET = 3_000_000;
+
+type ResendAttachmentMeta = {
+  filename?: string | null;
+  content_type?: string | null;
+  size?: number | null;
+};
+
+/** A PDF carried through step state, keyed back to its Resend attachment. */
+export type InlinePdfAttachment = PdfAttachment & { id: string };
+
+/**
+ * Whether an email's PDFs are small enough to pass through step state. An
+ * attachment with no reported size is treated as too big — the S3 fallback is
+ * always correct, so unknowns resolve that way.
+ */
+export function pdfsFitInlineBudget(
+  attachments: ResendAttachmentMeta[],
+): boolean {
+  const pdfs = attachments.filter((a) =>
+    isPdf({ filename: a.filename, contentType: a.content_type }),
+  );
+  if (pdfs.length === 0) return true;
+  if (pdfs.some((a) => typeof a.size !== "number")) return false;
+
+  const total = pdfs.reduce((sum, a) => sum + (a.size ?? 0), 0);
+  return total <= INLINE_ATTACHMENT_BUDGET;
+}
+
+/**
  * Fetch an inbound email's PDF attachments straight from Resend, as base64
  * `document` blocks. Used by the triage gate, which runs before we upload to
  * S3 or write a NotificationSource and so cannot use `fetchPdfAttachments`
+ *
+ * `complete` is false when any PDF failed to download. Callers must not pass a
+ * partial set downstream — the missing PDF would be silently absent, where the
+ * S3 fallback might still have retrieved it.
  */
 export async function fetchPdfAttachmentsFromResend(
   emailId: string,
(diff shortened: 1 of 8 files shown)


src/features/inbox/utils.ts
```diff
@@ -12,10 +13,30 @@ const isPdf = (a: {
       a.filename?.toLowerCase().endsWith(".pdf"),
   );
 
+/**
+ * Ceiling on the total base64 length of an email's PDFs for them to be carried
+ * through Inngest step state instead of re-fetched from S3. The encoded string
+ * is what rides in the step output, which Inngest caps at 4MiB (4,194,304) —
+ * the remainder is headroom for the rest of the step's JSON.
+ */
+export const INLINE_ATTACHMENT_BUDGET = 4_000_000;
```
buggyhuy, line 22: **Why 4,000,000 and why measured this way.**

Inngest caps a step's returned payload at 4MiB (4,194,304 bytes). The thing that actually occupies that budget is the base64 *string*, and base64's alphabet (`A-Z a-z 0-9 + / =`) is escape-free and single-byte in UTF-8 JSON — so `base64.length` is exactly the number of bytes this contributes. No conversion, no margin. The ~194KB gap is concrete headroom for the rest of the step's JSON: `kind`, `reasonWhy`, each attachment's `id`/`filename`, and punctuation.

The first revision of this PR summed Resend's per-attachment `size` instead and applied a ~4/3 expansion guess, because that field is available before downloading. That was the wrong instinct: `pdfsFitInlineBudget` runs *after* `fetchPdfAttachmentsFromResend`, so there was never anything to save by deciding early — the gate needs the PDFs either way. Estimating bought nothing and cost a dependency on an undocumented unit (Resend's API reference for `size` is an unfilled template artifact). Please don't reintroduce it.

What this constant does **not** cover: the 32MiB run-state ceiling. A budget-sized payload re-rides with the accumulated state on every later step invocation. That's the number to watch if latency regresses — and this constant is the only thing that needs to change.

src/features/inbox/utils.ts
```diff
@@ -57,7 +79,8 @@ export async function fetchPdfAttachmentsFromResend(
     }),
   );
 
-  return results.filter((a) => a !== null);
+  const downloaded = results.filter((a) => a !== null);
+  return { pdfs: downloaded, complete: downloaded.length === pdfs.length };
```
buggyhuy, line 83: `complete` compares against the count of **filtered PDFs**, not all attachments — a non-PDF that fails elsewhere can't false-negative it.

The reason this flag exists at all: before this PR, a failed download was silently dropped (`return null`, filtered out) and the agents just saw fewer documents. That was tolerable when every step independently fetched from S3 — a transient failure in one step didn't poison the others. Once we inline, a partial set becomes *sticky*: it's memoized into step state and every downstream agent inherits the same gap, with no second chance from S3. So the caller refuses to inline anything unless the set is whole.

src/inngest/functions/process-inbox-email.ts
```diff
@@ -54,20 +57,29 @@ export const processInboxEmail = inngest.createFunction(
...
-          subject: email.subject,
-          markdown: markdown ?? "",
-        },
-        pdfAttachments,
-      );
-    });
+    //
+    // The PDFs downloaded here are carried forward to every later step when
+    // they fit the budget, so nothing below re-fetches them from S3.
+    const { kind, attachments: inlinedPdfs } = await step.run(
```
buggyhuy, line 63: **The whole optimization is this one step's return value.**

`classify-email` was already paying for every PDF download — the gate deliberately reads attachments because "an email's real content is often only in the attachment" — and then threw the buffers away at the step boundary by returning only `{ kind }`. It now hands them forward.

`attachments === null` is the single fallback signal for all five consumers below. One flag, decided once, interpreted identically everywhere; no per-step size logic and no partial states to reason about. Both gates are conservative in the same direction: over budget → `null`, incomplete download → `null`, and `null` means every step behaves exactly as it did before this PR. The optimization can only ever remove work, never change what an agent sees.

The `not_relevant` early return below is unaffected — we return before anything is used.

src/inngest/functions/process-inbox-email.ts
```diff
@@ -79,29 +91,41 @@ export const processInboxEmail = inngest.createFunction(
...
         if (!email.attachments?.length) return [];
 
+        const inlinedById = new Map(
+          (inlinedPdfs ?? []).map((a) => [a.id, a.base64]),
+        );
+
         const { Resend } = await import("resend");
         const resend = new Resend(process.env.RESEND_API_KEY);
 
+        const loadBuffer = async (att: { id: string }) => {
```
buggyhuy, line 101: Extracted rather than branching inline because this `.map()` body now has two buffer sources and already does md5 + dedupe + PUT; nesting the Resend path inside it pushed the interesting part too far right to read.

Note this step iterates **all** attachments while the inlined map only ever contains PDFs — non-PDF attachments still take the Resend path. That's expected, not a gap: only PDFs are ever sent to the agents, so only PDFs are worth carrying through step state.

Matching is on Resend's attachment `id`, not `filename` — a single email can carry two attachments with the same filename, and keying on it would hand the wrong bytes to `uploadBufferToS3`.

buggyhuy (commented):
Context on the parts most likely to be misread later.

src/inngest/functions/process-inbox-email.ts
0xcad: +1, this is clean

0xcad (approved):

perrydev17 (approved):

0xcad (approved):

PR #171, opened 2026-07-21
[VW-366] Add ability to update version of affected assets

perrydev17:
ticket: https://northeastern-patch.atlassian.net/browse/VW-366

## Summary
- added a DropdownCell to the Affect Assets table, allowing users to update the version directly. It also supports the new Unsure value
- Assets are updated dynamically and moved to the another bucket when the version changes, and a toast message informs the user of the change

https://github.com/user-attachments/assets/2bcc97a9-36fe-40e4-91f8-1de2dfeaba09

## Summary by CodeRabbit

* **New Features**
  * Added support for recording uncertain or unknown asset versions via `versionStatus` (including “UNSURE”).
  * Added interactive version selection for affected assets with suggested versions and “Unknown”/“Not sure” options.
  * Asset updates now derive device-group version information (from existing vendor/product plus incoming version details) and record device-group history when applicable.
  * Notifications now include assets tied to unknown-version device groups in the “Needs information” bucket.
  * Added version lookups for vendor/product to power per-asset version suggestions.

### DIFF
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -195,10 +195,10 @@ model Connection {
 }
 
 enum VersionStatus {
-  UNKNOWN
+  UNKNOWN             // No answer exists, nobody has looked at this device's version
   NOT_APPLICABLE
   KNOWN
-  UNSURE
+  UNSURE              // A user looked and explicitly couldn't determine the version
 }
 
 // VERS version-range schemes (https://github.com/package-url/vers-spec)
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -198,6 +198,7 @@ enum VersionStatus {
   UNKNOWN
   NOT_APPLICABLE
   KNOWN
+  UNSURE
 }
 
 // VERS version-range schemes (https://github.com/package-url/vers-spec)
--- a/scripts/seed-notifications.ts
+++ b/scripts/seed-notifications.ts
@@ -374,7 +374,8 @@ const DESERIALIZATION_MATCHINGS: Array<{
 // Assets, each bucketed into a DeviceGroup keyed by exact version.
 const DESERIALIZATION_ASSETS: Array<{
   product: string;
-  version: string;
+  version?: string;
+  versionStatus?: "UNSURE";
   ip: string;
   hostname: string;
   serialNumber: string;
@@ -453,6 +454,23 @@ const DESERIALIZATION_ASSETS: Array<{
     role: "SPECT/CT Console",
     networkSegment: "IMAGING-NUCMED",
   },
+  {
+    product: "MAGNETOM Family",
+    ip: "10.60.0.13",
+    hostname: "mri-magnetom-03",
+    serialNumber: "MAGNETOM-UNKNOWN-001",
+    role: "MRI Scanner Console",
+    networkSegment: "IMAGING-MRI",
+  },
+  {
+    product: "MAGNETOM Family",
+    versionStatus: "UNSURE",
+    ip: "10.60.0.14",
+    hostname: "mri-magnetom-04",
+    serialNumber: "MAGNETOM-UNKNOWN-001",
+    role: "MRI Scanner Console",
+    networkSegment: "IMAGING-MRI",
+  },
 ];
 
 async function upsertDeserializationMatching(spec: {
@@ -477,16 +495,22 @@ async function upsertDeserializationMatching(spec: {
   );
 }
 
-async function upsertDeviceGroupForAsset(product: string, version: string) {
+async function upsertDeviceGroupForAsset(
+  product: string,
+  version?: string,
+  versionStatus: VersionStatus = version
+    ? VersionStatus.KNOWN
+    : VersionStatus.UNKNOWN,
+) {
   const vendor = await upsertVendor(DESERIALIZATION_VENDOR);
   const productRec = await upsertProduct(product);
-  const versionRec = await upsertVersion(version);
+  const versionRec = version ? await upsertVersion(version) : null;
 
   const identity = {
     vendorId: vendor.id,
     productId: productRec.id,
-    versionId: versionRec.id,
-    versionStatus: VersionStatus.KNOWN,
+    versionId: versionRec?.id ?? null,
+    versionStatus: VersionStatus,
   };
 
   return (
@@ -562,8 +586,14 @@ async function seedDeserializationScenario(userId: string) {
     const deviceGroup = await upsertDeviceGroupForAsset(
       spec.product,
       spec.version,
+      spec.versionStatus === "UNSURE" ? VersionStatus.UNSURE : undefined,
     );
-    const { product: _product, version: _version, ...assetFields } = spec;
+    const {
+      product: _product,
+      version: _version,
+      versionStatus: _versionStatus,
+      ...assetFields
+    } = spec;
     assets.push(
       await prisma.asset.create({
         data: {
--- a/scripts/seed-notifications.ts
+++ b/scripts/seed-notifications.ts
@@ -506,16 +506,17 @@ async function upsertDeviceGroupForAsset(
   const productRec = await upsertProduct(product);
   const versionRec = version ? await upsertVersion(version) : null;
 
-  const identity = {
-    vendorId: vendor.id,
-    productId: productRec.id,
-    versionId: versionRec?.id ?? null,
-    versionStatus: VersionStatus,
-  };
+  const vendorId = vendor.id;
+  const productId = productRec.id;
+  const versionId = versionRec?.id ?? null;
 
   return (
-    (await prisma.deviceGroup.findFirst({ where: identity })) ??
-    (await prisma.deviceGroup.create({ data: identity }))
+    (await prisma.deviceGroup.findFirst({
+      where: { vendorId, productId, versionId, versionStatus },
+    })) ??
+    (await prisma.deviceGroup.create({
+      data: { vendorId, productId, versionId, versionStatus },
+    }))
   );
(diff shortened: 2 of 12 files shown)


prisma/schema.prisma
0xcad, line 1: What's the difference between "UNSURE" and "UNKNOWN"?

Here's my guess, is this accurate?
* Unknown -- Our platform does not know the version
* Unsure -- Our platform, and the user, does not know the version

scripts/seed-notifications.ts
```diff
@@ -453,6 +454,23 @@ const DESERIALIZATION_ASSETS: Array<{
...
+  },
+  {
+    product: "MAGNETOM Family",
+    versionStatus: "UNSURE",
+    ip: "10.60.0.14",
+    hostname: "mri-magnetom-04",
+    serialNumber: "MAGNETOM-UNKNOWN-001",
+    role: "MRI Scanner Console",
+    networkSegment: "IMAGING-MRI",
+  },
```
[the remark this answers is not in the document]
0xcad: @perrydev17 looks accurate

src/features/assets/server/routers.ts
```diff
@@ -607,18 +608,48 @@ export const assetsRouter = createTRPCRouter({
...
           utilization: utilization as Prisma.InputJsonValue | undefined,
         },
         include: assetInclude,
       });
+      if (deviceGroupId) {
+        await prisma.deviceGroupHistory.create({
+          data: { deviceGroupId, assetId: id },
+        });
+      }
+      return updated;
```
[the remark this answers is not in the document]
0xcad: @perrydev17 this also looks accurate to me

src/features/assets/types.ts
```diff
@@ -44,6 +44,14 @@ export const assetInputSchema = z.object({
   location: locationSchema.optional(),
   status: assetStatusSchema.nullish(),
   utilization: assetUtilizationSchema.nullish(),
+  version: z
+    .string()
+    .trim()
+    .min(1)
+    .max(64)
+    .regex(/^[\w.\- +:|]+$/)
```
0xcad: This regex seems potentially overly restrictive when we're not sure what version naming schemes we might see

I would just get rid of it

Currently this would reject versions like:
* `VBFOO (legacy edition)`
* `1.2.3#abc7f2e`
* `1.5.0/beta`
* `1.0.0(custom)`

^Not sure if any of those are real, but it's nice to have flexibility

src/features/inbox/components/DropdownCell.tsx
```diff
@@ -0,0 +1,77 @@
...
+      <SelectTrigger className="h-7 text-sm w-36">
+        <SelectValue>{currentDisplayLabel}</SelectValue>
+      </SelectTrigger>
+      <SelectContent>
+        {options.map((o) => (
+          <SelectItem key={o} value={o}>
+            {o}
+          </SelectItem>
+        ))}
+        <SelectSeparator />
```
0xcad: Change requested: I would put "Unknown" and "Unsure" first, above options, since `options` could be long

src/features/inbox/components/DropdownCell.tsx
```diff
@@ -0,0 +1,77 @@
...
+      : selected === UNSURE_VALUE
+        ? "Not sure"
+        : selected;
+
+  return (
+    <Select
+      value={selected}
+      disabled={isPending}
+      onValueChange={(v) => handleDropDownValueChange(v)}
+    >
```
0xcad, line 55: Can you @perrydev17  create a ticket, just so we're tracking it, to allow an input field into this select? Not necessary for this PR though

Intended behavior:
* Can use input field which functions as a search, narrowing down options
* User can also just leave their input field as-is, which is equivalent to a write-in response
* "Unknown" and "Not Sure" are always visible

src/features/inbox/components/DropdownCell.tsx
```diff
@@ -0,0 +1,77 @@
...
+  Select,
+  SelectContent,
+  SelectItem,
+  SelectSeparator,
+  SelectTrigger,
+  SelectValue,
+} from "@/components/ui/select";
+
+const UNKNOWN_VALUE = "Unknown";
+const UNSURE_VALUE = "Not sure";
```
0xcad, line 14: Change requested

Still not quite sure what the difference between "Unknown" and "Not sure" is, here's [my guess](https://github.com/PATCH-UPGRADE/viper/pull/171/changes#r3631076600) again

If that's accurate, then I vote the user should not have to see the "Unknown" option at all. Either the user knows what the version is, in which case they select the correct option, or they don't know what it is, and they hit "Not sure".

"Unknown" and "Not sure" seems too confusing/redundant, to both be user choices

src/features/inbox/components/notification-affected-assets-tab.tsx
```diff
@@ -49,6 +55,24 @@ function MatchingAssetTable({
...
+
+  const suggestedVersion =
+    displayName(matching.version) ?? matching.versionRange;
+
+  const options = useMemo(() => {
+    const names = (knownVersions ?? []).map((v) => v.canonicalDisplayName);
+    return suggestedVersion && !names.includes(suggestedVersion)
+      ? [suggestedVersion, ...names]
+      : names;
+  }, [knownVersions, suggestedVersion]);
```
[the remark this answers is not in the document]
0xcad: @perrydev17 Change requested, I agree with this. Your example video shows a complex version range. I think eventually we may want to support version ranges for device groups -- I could imagine a usecase where the user isn't sure what the version is, but they know it's at least X, and it's not yet Y -- but right now our db isn't set up for that

src/features/inbox/components/notification-affected-assets-tab.tsx
```diff
@@ -49,6 +55,24 @@ function MatchingAssetTable({
   const [rows, setRows] = useState<ResolvedDeviceGroupAsset[]>([]);
   const appendedPages = useRef(new Set<number>());
 
+  const matching = group.deviceGroupMatching;
+  const { data: knownVersions } = useVersionForVendorProduct({
+    vendorId: matching.vendorId,
+    productId: matching.productId ?? "",
+  });
```
0xcad, line 66: This calls the hook for each DGM, in each bucket, correct? In practice I think that produces cases where you call this endpoint with the same vendor/product maybe 2-3 times extra, unnecessarily (consider case where DGM with same vendor/product appears in both AFFECTED and NOT_AFFECTED, or two DGM with same vendor/product are both AFFECTED but with different version ranges). I think you could perhaps make this more efficient by instead:

* Iterating over all DGM's to build a set of vendor/product tuples
* Calling `useVersionForVendorProduct` once on each vendor/product tuple
* Saving the results, keyed to (vendorId, productId)
* Pass that dictionary into MatchingAssetTable
* Get the version options here by looking up vendorId / productId

But also premature optimization is the root of all evil so this should be fine as-is, no change needed

0xcad (changes requested):
Changes also put in comments where relevant 

* Clarification on Unsure/Unknown needed (leave a comment in prisma schema?)
* Some code rabbit suggestions
* get rid of version regex
* Get rid of "Unknown" as option in select dropdown, move "Not sure" to be above options
  * Create a ticket for input field in version dropdown

prisma/schema.prisma
perrydev17: Unknown is the default/absence state, (the original `-`), our platform does not know the version, nobody has updated it yet. (which also mean yes, we should remove unknown from the dropdown)

Unsure, user active response. A user looked at this specific device, tried to determine its version, but couldn't

src/features/assets/types.ts
perrydev17: removed

src/features/inbox/components/DropdownCell.tsx
perrydev17: updated with remove Unknown, put Not sure first above options
Screenshot 2026-07-22 at 12 49 00 PM

src/features/inbox/components/DropdownCell.tsx
perrydev17: sure, https://northeastern-patch.atlassian.net/browse/VW-384 (only Not sure is always visible), removing Unknown in this pr

src/features/inbox/components/DropdownCell.tsx
perrydev17: yes, I removed Unknown from the options. User can update the version from UNKNOWN to either a specific version number or Not sure, but not the other way around.

src/features/inbox/components/notification-affected-assets-tab.tsx
perrydev17: useVersionForVendorProduct currently dedup by React Query. For example, if the same matching appears in two buckets, or two different DGM rows happen to share the same vendor/product, they will share the same React Query cache because they have the same vendorId, productId.  

I added a todo comment in the code as a follow up optimization

src/features/assets/server/routers.ts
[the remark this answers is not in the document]
perrydev17: updated

scripts/seed-notifications.ts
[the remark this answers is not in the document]
perrydev17: updated

src/features/inbox/components/notification-affected-assets-tab.tsx
0xcad: > useVersionForVendorProduct currently dedup by React Query

Oh, I wasn't following that but that's great, good to know ty!

0xcad (approved):

PR #176, opened 2026-07-22
[VW-0] Resolve npm audit vulnerabilities (43 → 7)

buggyhuy:
## Summary

Clears the `npm audit` wall that shows up on every `npm install`. Audit goes from 43 vulnerabilities (5 critical, 13 high, 17 moderate, 8 low) down to 7, and every critical, high, and moderate advisory is resolved. Dependency-only change — `package.json` and `package-lock.json`, no source touched.

Fixes were applied by risk rather than by running `npm audit fix --force` (I just let ClaudeCode do the fixing here). Force wanted to downgrade Next.js from 15 to 9.3.3 to satisfy `sharp` and `postcss`, so those two are pinned through `overrides` instead. The rest: the vitest tree moves to 4.1.10 (clears the critical `@vitest/browser` file-access gate bypass), `adm-zip` is overridden to 0.6.0 (it only reaches us through the dev-only `inngest-cli`), `nodemailer` goes 7 → 9 (`src/lib/mail.ts` only uses `createTransport`/`sendMail`, stable across those majors), and `@scalar/api-reference-react` goes to 0.9.59, whose only consumer is the OpenAPI UI page.

**The 7 remaining lows are left in place on purpose.** They are all one advisory — `@ai-sdk/provider-utils` uncontrolled resource consumption ([GHSA-866g-f22w-33x8](https://github.com/advisories/GHSA-866g-f22w-33x8)). The only fix npm offers is a major upgrade of the whole Vercel AI SDK (v5 → v6: `ai@6.x`, `@ai-sdk/*@4.x`), which changes the chat-streaming surface we depend on — `useChat` and the LangGraph stream-bridge. That is a real migration with test risk, not a lockfile bump, so it belongs in its own ticket rather than riding along with dependency hygiene.

## Testing

- `npm audit`: 43 → 7 (0 critical / 0 high / 0 moderate; the 7 low are the deferred AI SDK advisory above).
- `npm run build` passes, including the `@scalar` OpenAPI UI page that the `api-reference-react` bump touches.
- Dev server (`:3300`) and Inngest (`:8588`) start clean — Inngest confirms `adm-zip@0.6.0` doesn't break the CLI. Smoke-tested against the running app.
- Reproduce: `npm install && npm audit` (expect 7 low), then `npm run build`.

## Summary by CodeRabbit

* **Chores**
  * Updated application and development dependencies to newer versions, including tooling and email/API libraries.
  * Expanded dependency-resolution safeguards by pinning additional package versions for improved stability and compatibility.
  * Reduced unnecessary runtime module inclusion by converting several imports to type-only usage (no user-facing behavior changes).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -48,7 +48,7 @@
         "@radix-ui/react-toggle": "^1.1.10",
         "@radix-ui/react-toggle-group": "^1.1.11",
         "@radix-ui/react-tooltip": "^1.2.8",
-        "@scalar/api-reference-react": "^0.8.6",
+        "@scalar/api-reference-react": "^0.9.59",
         "@sentry/nextjs": "^10.18.0",
         "@sparticuz/chromium": "^149.0.0",
         "@tanstack/react-query": "^5.90.2",
@@ -73,7 +73,7 @@
         "ms": "^2.1.3",
         "next": "^15.5.7",
         "next-themes": "^0.4.6",
-        "nodemailer": "^7.0.13",
+        "nodemailer": "^9.0.3",
         "nuqs": "^2.7.1",
         "playwright": "^1.61.1",
         "playwright-core": "^1.61.1",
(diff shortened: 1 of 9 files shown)


0xcad (approved):

perrydev17 (approved):
Thank you for taking care of this.

PR #180, opened 2026-07-23
[VW-372387] Add Question agent: model, prompt (1/3)

perrydev17:
ticket: https://northeastern-patch.atlassian.net/browse/VW-387

## Summary
- This is the first pr of the full Question agent feature, it covers adding a new Question model, creating the Question agent (mirror Vex's approach). The next pr will cover when user answer a question, note creation if necessary, status updates, triage re-run, followup questions.  

## Out of scope
- API handling when user provide an answer, note creation, status updates, triage re-run and generate followup questions
- Frontend UI

local verification without API and UI

```
stdout | src/features/inbox/agent/question/__tests__/question.test.ts > generate Questions > creates questions for the seeded UNDER_INVESTIGATION issue
Issue vuln  cmrl6o8qf003kj81wcp9we5ip

stdout | src/features/inbox/agent/question/__tests__/question.test.ts > generate Questions > creates questions for the seeded UNDER_INVESTIGATION issue
notification  cmrnoty850009j8j3lz8b3tpt

stdout | src/features/inbox/agent/question/__tests__/question.test.ts > generate Questions > creates questions for the seeded UNDER_INVESTIGATION issue
Notification Vuln  [ 'cmrl6o8qf003kj81wcp9we5ip' ]

stdout | src/features/inbox/agent/question/__tests__/question.test.ts > generate Questions > creates questions for the seeded UNDER_INVESTIGATION issue
Questions created:  [
  {
    "id": "cmry3v13o0001j8ybpc6k62q2",
    "issueId": "cmrl6o8qn003lj81w46xtk1yp",
    "notificationId": "cmrnoty850009j8j3lz8b3tpt",
    "title": "Is the local password store on the remaining affected syngo.plaza asset(s) accessible over the network, or only via physical/local console access?",
    "reasonWhy": "The advisory does not specify whether exploiting the insecure password encryption requires network-level access to the password store or physical console access. With pacs-syngo-02 confirmed as migrated to SSO/AD, the risk hinges on how the local password store on the other affected asset(s) is exposed.",
    "suggestedAnswers": [
      "Network-accessible (e.g. shared drive, API, or remote desktop)",
      "Local/physical console access only",
      "Local network access only (not internet-facing)",
      "Unknown – network exposure has not been assessed"
    ],
    "status": "PENDING",
    "answer": null,
    "answeredByUserId": null,
    "answeredAt": null,
    "resultingNoteId": null,
    "createdAt": "2026-07-23T22:51:09.779Z",
    "updatedAt": "2026-07-23T22:51:09.779Z"
  }
]

 ✓ src/features/inbox/agent/question/__tests__/question.test.ts (6 tests) 25ms
   ✓ generate Questions (1)
     ✓ creates questions for the seeded UNDER_INVESTIGATION issue 23ms
   ✓ planQuestionWrites (5)
     ✓ creates an op for a valid, in-scope issue with a complete question 0ms
     ✓ skips an entry with an empty or whitespace-only title 0ms
     ✓ skips an entry with fewer than two suggested answers 0ms
     ✓ treats an omitted issue as no question - not an error 0ms
     ✓ handles multiple valid issues in one result 0ms

 Test Files  1 passed (1)
      Tests  6 passed (6)
   Start at  19:07:49
   Duration  524ms (transform 79ms, setup 59ms, import 144ms, tests 25ms, environment 196ms)```


<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * Added automated follow-up question generation for inbox issues still under investigation.
  * Questions now include a title, explanation, and suggested answers, and can be linked to notifications, issues, users, and notes.
  * Added question status tracking, including pending, answered, dismissed, and unsure states.
  * Added safeguards to prevent incomplete or unsupported questions from being saved.
* **Bug Fixes**
  * Improved notification seeding for advisories affecting device groups.
* **Tests**
  * Added coverage for valid, invalid, missing, and multiple generated questions.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -53,6 +53,7 @@ model User {
   ticketSeen             TicketSeen[]
   ticketActivities       TicketActivity[]
   notes                  Note[]
+  answeredQuestions Question[]
 
   @@unique([email])
   @@index([departmentId])
@@ -657,6 +658,7 @@ model Issue {
   issueRemediations  IssueRemediation[]
   workOrderTickets      WorkOrderTicket[]     @relation("WorkOrderTicketIssues")
   filters           EntityFilter[]
+  questions Question[]
 
   createdAt DateTime @default(now())
   updatedAt DateTime @updatedAt
@@ -693,6 +695,7 @@ model Note {
 
   targetModel ScopeTargetModel?
   instanceId  String? // weak reference — app validates this id exists in targetModel's table
+  answeredQuestion Question?
 
   createdAt DateTime @default(now())
   updatedAt DateTime @updatedAt
@@ -1269,6 +1272,7 @@ model Notification {
   matchFeedback   MatchFeedback[]
   mitigationPlans MitigationPlan[]
   workOrderTickets WorkOrderTicket[]
+  questions Question[]
 
   createdAt DateTime @default(now())
   updatedAt DateTime @updatedAt
@@ -1468,7 +1472,7 @@ enum CorrectionTargetType {
 }
 
 model FieldCorrection {  
-  id  String @id @default(uuid())
+  id  String @id @default(cuid())
   targetType CorrectionTargetType
   targetId  String
   field   String
@@ -1485,4 +1489,43 @@ model FieldCorrection {
   @@index([userId])
   @@index([field, createdAt])
   @@map("field_correction")
+}
+
+// ─── Question ────────────────────────────────────────────────────────────
+
+enum QuestionStatus {
+  PENDING   // default, not answered by User
+  ANSWERED
+  DISMISSED
+  UNSURE
+}
+
+model Question {
+  id  String @id @default(cuid())
+  issueId String
+  issue  Issue @relation(fields: [issueId], references: [id], onDelete: Cascade)
+
+  notificationId String
+  notification Notification @relation(fields: [notificationId], references: [id], onDelete: Cascade)
+
+  title String
+  reasonWhy String @db.Text
+  suggestedAnswers String[] @default([])
+  
+  status QuestionStatus @default(PENDING)
+  answer String?  @db.Text
+  answeredByUserId  String?
+  answeredBy  User? @relation(fields: [answeredByUserId], references: [id], onDelete: SetNull)
+  answeredAt DateTime?
+
+  resultingNoteId String? @unique   // create note after user answer
+  resultingNote Note? @relation(fields: [resultingNoteId], references: [id], onDelete: Cascade)
+
+  createdAt DateTime @default(now())
+  updatedAt DateTime @updatedAt
+
+  @@index([issueId])
+  @@index([notificationId])
+  @@index([status])
+  @@map("question")
 }
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -1519,7 +1519,7 @@ model Question {
   answeredAt DateTime?
 
   resultingNoteId String? @unique   // create note after user answer
-  resultingNote Note? @relation(fields: [resultingNoteId], references: [id], onDelete: Cascade)
+  resultingNote Note? @relation(fields: [resultingNoteId], references: [id], onDelete: SetNull)
 
   createdAt DateTime @default(now())
   updatedAt DateTime @updatedAt
--- a/src/features/inbox/agent/question/context.ts
+++ b/src/features/inbox/agent/question/context.ts
@@ -1,0 +1,204 @@
+import "server-only";
+import { getRelevantNotes } from "@/features/notes/server/get-relevant-notes";
+import prisma from "@/lib/db";
+import {
+  deviceGroupWhereForMatching,
+  type MatchingLike,
+  matchingAppliesToDeviceGroup,
+} from "@/lib/device-matching";
+import { deviceGroupMatchingLabel, type NoteRow } from "@/lib/markdown";
+
+export type QuestionIssueContext = { issueId: string; vulnerabilityId: string };
+
+export type QuestionContext = {
+  notificationId: string;
+  markdown: string;
+  issues: QuestionIssueContext[];
+};
+
+type MatchingWithRefs = MatchingLike & {
+  id: string;
+  vendor?: { canonicalDisplayName: string } | null;
+  product?: { canonicalDisplayName: string } | null;
+  version?: { canonicalDisplayName: string } | null;
+};
+
+function renderQuestionPrompt(args: {
+  vulnerabilities: Array<{
+    cveId: string | null;
+    severity: string;
+    description: string | null;
+  }>;
+  issueRenders: Array<{
+    issueId: string;
+    cve: string;
+    matching: MatchingWithRefs;
+    assetCount: number;
+    statusNotes: string | null;
+  }>;
+  notes: NoteRow[];
+}): string {
+  const sections: string[] = [];
+
+  sections.push(
+    "## Linked vulnerabilities\n\n" +
+      args.vulnerabilities
+        .map(
+          (v) =>
+            `### ${v.cveId ?? "unknown CVE"} - ${v.severity}\n${v.description ?? ""}`,
+        )
+        .join("\n\n"),
+  );
+
+  if (args.notes.length > 0) {
+    sections.push(
+      "## Notes (evidence)\n\n" + args.notes.map((n) => `${n.text}`).join("\n"),
+    );
+  }
+  sections.push(
+    "## Issues that need a question \n\n" +
+      "Each of these is currently UNDER INVESTIGATION. Draf a specific, answerable " +
+      " question for any you can meaningfully clarify. Omit an id to skip it. \n\n" +
+      args.issueRenders
+        .map((r) =>
+          [
+            `- Issue \ ${r.issueId}\ - ${r.cve}`,
+            `- Device: ${deviceGroupMatchingLabel(r.matching)}`,
+            ` - Assets affected: ${r.assetCount}`,
+            ` - Why this is under investigation: ${r.statusNotes ?? "(no reason recorded)"}`,
+          ].join("\n"),
+        )
+        .join("\n\n"),
+  );
+
+  return sections.join("\n\n");
+}
+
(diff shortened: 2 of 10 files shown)

### CODE REVIEW

File: prisma/schema.prisma [hunk 1]
```diff
@@ -1472,7 +1472,7 @@ enum CorrectionTargetType {
 }
 
 model FieldCorrection {  
-  id  String @id @default(uuid())
+  id  String @id @default(cuid())
```
0xcad, line 1475: +1, thanks for catching that

prisma/schema.prisma
```diff
@@ -1490,3 +1490,42 @@ model FieldCorrection {
...
+
+  notificationId String
+  notification Notification @relation(fields: [notificationId], references: [id], onDelete: Cascade)
+
+  title String
+  reasonWhy String @db.Text
+  suggestedAnswers String[] @default([])
+  
+  status QuestionStatus @default(PENDING)
+  answer String?  @db.Text
```
0xcad, line 1516: and presumably we'll enforce (via trpc?), that a question with an answer has QuestionStatus answered?

Idea: ability for user to answer, provide us more info, but still be unsure?

src/features/inbox/agent/question/context.ts
```diff
@@ -0,0 +1,204 @@
...
+  );
+
+  if (args.notes.length > 0) {
+    sections.push(
+      "## Notes (evidence)\n\n" + args.notes.map((n) => `${n.text}`).join("\n"),
+    );
+  }
+  sections.push(
+    "## Issues that need a question \n\n" +
+      "Each of these is currently UNDER INVESTIGATION. Draf a specific, answerable " +
```
0xcad: Draf -> "Draft"

src/features/inbox/agent/question/index.ts
```diff
@@ -0,0 +1,60 @@
...
+  type QuestionContext,
+  SYSTEM_PROMPT,
+} from "./context";
+import {
+  applyQuestionWrites,
+  type QuestionApplySummary,
+} from "./process_output";
+import { buildQuestionSchema, type QuestionResult } from "./schema";
+
+const MODEL = "claude-sonnet-4-6";
```
0xcad: I think we could probably do haiku for this agent but yours to consider

src/features/inbox/agent/question/index.ts
```diff
@@ -0,0 +1,60 @@
...
+  context: QuestionContext,
+): Promise<QuestionResult> {
+  const issueIds = context.issues.map((issue) => issue.issueId);
+  const schema = buildQuestionSchema(issueIds);
+
+  const model = new ChatAnthropic({
+    model: MODEL,
+    maxTokens: 4000,
+    thinking: { type: "enabled", budget_tokens: 2000 },
+  });
```
0xcad: If you go with Sonnet, then to get structured output working with thinking you have to use a tool. See how vex/index.ts does this: https://github.com/PATCH-UPGRADE/viper/blob/5db04b369ddac4c39f14d73351df8291d5a6ecdf/src/features/inbox/agent/vex/index.ts#L34-L40

If you ditch Sonnet for haiku, then structured output uses `withStructuredOutput`: https://github.com/PATCH-UPGRADE/viper/blob/5db04b369ddac4c39f14d73351df8291d5a6ecdf/src/features/inbox/agent/triage/index.ts#L80-L83

src/features/inbox/agent/question/context.ts
```diff
@@ -0,0 +1,204 @@
...
+}
+
+// ─── System prompt ───────────────────────────────────────────────────────────
+
+export const SYSTEM_PROMPT = `You are drafting clarifying questions for a hospital security engineer, for vulnerability issues a triage agent already marked "under investigation" because it lacked enough information to decide if the vulnerability is exploitable.
+For each issue, use the stated reason it's under investigation to write ONE specific, answerable question - never a vague "please provide more information." Include 2-6 short suggested answers a user could pick instead of typing. Ground every question in the evidence given; never invent facts about the device or vulnerability.
+Respond with ONLY a single JSON object - no prose, no markdown code fences, nothing before or after it. One key per issue id you're providing a question for, using the exact issue ids given. Omit any issue you don't have a good, specific question for. Each value must have exactly these fields: "title" (string), "reasonWhy" (string), "suggestedAnswers" (array of 2-6 short strings)
+Example shape: {"<issueId>": {"title":"..., "reasonWhy": "...", "suggestedAnswers":["...","..."]}}`;
```
0xcad: So in general, forcing an LLM to respond via a prompt with JSON is an anti-pattern. We want to use structured output, which forces the LLM to only allow a specific output

Thinking agents don't fully support structured output, so we have to create a tool the model calls, where the inputs to that tool is how it records its answer and ends it turn. If you go back to a model like haiku (I think that should be capable for this task) structured output becomes even easier

See my comments on question/index.ts:
https://github.com/PATCH-UPGRADE/viper/pull/180/changes#r3646887653

src/features/inbox/agent/question/index.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  let raw: unknown;
+  try {
+    raw = JSON.parse(extractJsonBlock(res.text));
+  } catch (e) {
+    return {};
+  }
+  const parsed = schema.safeParse(raw);
+
+  return parsed.success ? parsed.data : {};
```
0xcad: This is the kind of thing that structured output would handle for you

> return parsed.success ? parsed.data : {};

In general too, if parsed.success is False we want to fail, rather than accept this as an empty output

https://github.com/PATCH-UPGRADE/viper/blob/5db04b369ddac4c39f14d73351df8291d5a6ecdf/src/features/inbox/agent/vex/index.ts#L55-L60

We can think about a better pattern than just throwing an error (personally, I think that's pretty good -- if we throw an error, Inngest will retry the job to a maxcount of, iirc, 4 iterations, and the model has another attempt to get the types right). Regardless, I do believe that falling back to an empty json object with no logging is not the best soln -- but lmk if I'm mistaking anything in the code or behavior

src/features/inbox/agent/question/process_output.ts
```diff
@@ -0,0 +1,47 @@
...
+export function planQuestionWrites(
+  context: QuestionContext,
+  result: QuestionResult,
+): QuestionCreateOp[] {
+  const issueIds = new Set(context.issues.map((i) => i.issueId));
+  const ops: QuestionCreateOp[] = [];
+
+  for (const [issueId, value] of Object.entries(result)) {
+    if (!issueIds.has(issueId) || !value) continue;
+    if (!value.title.trim() || value.suggestedAnswers.length < 2) continue;
```
0xcad, line 21: Correct me if I'm wrong, should your Zod schema handle this?

```
const questionSchema = z.object({
  title: z.string(),
  reasonWhy: z.string(),
  suggestedAnswers: z.array(z.string()).min(2).max(6),
});
```

**.min(2)**?

src/inngest/functions/process-inbox-email.ts
```diff
@@ -396,7 +397,16 @@ export const processInboxEmail = inngest.createFunction(
...
 
-    // 11. Triage: assign priority, reason, and hospital impact
+    // 11. Generte questions for any Issue VEX just left UNDER_INVESTIGATION
+    const questionSummary = await step.run("generate-questions", async () => {
+      if (!notificationId || !vexSummary)
+        return { questionSkipped: true as const };
+      if ("vexSkipped" in vexSummary && vexSummary.vexSkipped)
+        return { questionSkipped: true as const };
+      return generateQuestionForNotification(notificationId);
+    });
```
0xcad, line 407: This is just a question, I have no idea how to do this, but is there a way in Inngest to run steps in parallel?

Because generate-questions step doesn't affect the output of triage-notification (unlike the previous steps, which build on each other)

Could be useful to run both of those in parallel so triage isn't blocked until questions get completed. Non-blocking of course

0xcad (changes requested):
Looks good, but a few structural changes related to structured output. I should link places in code of how that's implemented with sonnet (tool calls) or haiku (`.withStructuredOutput`). And smaller notes on typo, inngest steps

prisma/schema.prisma
perrydev17: yes, enforced via trpc, "ANSWERED" requires a non-empty answer. UNSURE can optionally carry an answer

src/features/inbox/agent/question/context.ts
perrydev17: fixed

src/features/inbox/agent/question/index.ts
perrydev17: updated to haiku

src/features/inbox/agent/question/index.ts
perrydev17: updated to haiku with StructuredOutput

src/features/inbox/agent/question/context.ts
perrydev17: fixed, removed all the prompt with JSON, using StructuredOutput instead

src/features/inbox/agent/question/process_output.ts
perrydev17: yes it does, I try to use the same approach from planVexWrites to have a re-check even though the schema already handle it

src/inngest/functions/process-inbox-email.ts
perrydev17: right, referring to this https://www.inngest.com/docs/examples/durable-endpoints#running-steps-in-parallel, it is possible, so I am going to try with Promise.all([step.run('generate-questions'), step.run('triage-notification)] in next pr

src/features/inbox/agent/question/index.ts
perrydev17: you are right on the 4 retries (https://www.inngest.com/docs/features/inngest-functions/error-retries/retries), and fail it to let it retry is a better handling

0xcad (approved):

PR #182, opened 2026-07-24
[VW-0] Making Playwright imports Dynamic

markglose-bc:
Was getting the following 500 error on vercel deployments
image

Moving (playwright-core, @sparticuz/chromium) to a dynamic import in `launchBrowser()` resolves this 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved browser launching reliability across different deployment environments.
  * Enhanced compatibility for hosted deployments using the Vercel platform.

### DIFF
--- a/src/features/integrations/teamplay-fleet/capture.ts
+++ b/src/features/integrations/teamplay-fleet/capture.ts
@@ -1,6 +1,3 @@
-import sparticuzChromium from "@sparticuz/chromium";
-import { chromium } from "playwright-core";
-
 export interface SessionLoginConfig {
   welcomeUrl: string;
   cookieBannerAcceptSelector?: string; // there is a cookie setting overlay on the very first login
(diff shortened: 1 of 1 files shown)


perrydev17 (approved):
interesting on this error, it was ok for a few days and suddenly broke?

markglose-bc:
> interesting on this error, it was ok for a few days and suddenly broke?

I looked into this a bit more, VW-350 (creating work orders on Fleet) was the primary cause.

`capture.ts` has imported `playwright-core` at module scope, but it was a disconnected subtree — only Inngest/API routes reached it, and those have the `outputFileTracingIncludes` entry that ships playwright's files.

VW-350 wired the `tracking` tRPC router to `teamplay-fleet/tracking` → `config` (`export const FLEET = new IntegrationSessionClient(...)`) → `capture` → `playwright-core`. Since `tracking` is mounted in the app router, and every page imports the app router via `@/trpc/server` (prefetch/HydrateClient), playwright got added into every page's serverless bundle.

At runtime playwright loads `browsers.json` via a computed `require(path.join(...))`, which Vercel's file tracer can't follow, so it was never copied into the page functions → `Cannot find module browsers.json` → 500 at module load

PR #183, opened 2026-07-25
[VW-388] Question Agent API, Create Note, Generate Followup questions (2/3)

perrydev17:
ticket : https://northeastern-patch.atlassian.net/browse/VW-388
## Summary
- new API to handle user actions with Question
- add handling for creating note, generate followup questions, updated prompt

How to verify locally without UI
- I created a temp test script and hard code to notification with "syngo.plaza"
- confirm question was generated
- updated answer(without providing actual info) and status from local DB directly
- confirm Inngest re-run and it generates a follow up question

inngest_question

## Summary by CodeRabbit

* **New Features**
  * Added support for reviewing, answering, dismissing, or marking issue questions as unsure.
  * Added question history views by notification and issue.
  * Added automatic follow-up questions using previous answers to avoid repetition.
  * Added automatic issue re-evaluation after an answer, including updated prioritization and triage.
  * Notification processing now performs question generation, triage, and mitigation planning concurrently.

### DIFF
--- a/src/features/inbox/agent/question/context.ts
+++ b/src/features/inbox/agent/question/context.ts
@@ -105,13 +105,13 @@ export async function gatherQuestionContext(
       },
     },
   });
-
+  console.log("gatherQuestionContext notification ", notification);
   if (!notification) return null;
 
   const vulnerabilities = notification.vulnerabilities.map(
     (m) => m.vulnerability,
   );
-
+  console.log("gatherQuestionContext vulnerabilities ", vulnerabilities);
   if (vulnerabilities.length === 0) return null;
 
   const matchingsById = new Map<string, MatchingWithRefs>();
@@ -119,7 +119,7 @@ export async function gatherQuestionContext(
     for (const dgm of v.deviceGroupMatchings) matchingsById.set(dgm.id, dgm);
   }
   const matchings = [...matchingsById.values()];
-
+  console.log("gatherQuestionContext matchings ", matchings);
   const candidateGroups =
     matchings.length > 0
       ? await prisma.deviceGroup.findMany({
@@ -154,9 +154,11 @@ export async function gatherQuestionContext(
     statusNotes: string | null;
   };
   const issueRenders: IssueRender[] = [];
-
+  console.log("gatherQuestionContext-----------------------------");
   for (const v of vulnerabilities) {
+    console.log("gatherQuestionContext v ", v);
     for (const issue of v.issues) {
+      console.log("gatherQuestionContext issue ", issue);
       if (!issue.deviceGroupMatchingId) continue;
       const matching = matchingsById.get(issue.deviceGroupMatchingId);
 
@@ -177,7 +179,7 @@ export async function gatherQuestionContext(
       });
     }
   }
-
+  console.log("gatherQuestionContext issues ", issues);
   if (issues.length === 0) return null;
 
   const assetIds = [
@@ -188,12 +190,13 @@ export async function gatherQuestionContext(
     deviceGroupMatchingIds: matchings.map((m) => m.id),
     assetIds,
   });
-
+  console.log("gatherQuestionContext notes ", notes);
   const markdown = renderQuestionPrompt({
     vulnerabilities,
     issueRenders,
     notes,
   });
+  console.log("gatherQuestionContext markdown ", markdown);
   return { notificationId, markdown, issues };
 }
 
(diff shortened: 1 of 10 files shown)


src/inngest/functions/reevaluate-issue-on-answer.ts
```diff
@@ -0,0 +1,71 @@
...
+          data: {
+            priority: result.priority,
+            priorityReasonWhy: result.priorityReasonWhy,
+            hospitalImpact: result.hospitalImpact,
+          },
+        });
+      });
+    } else {
+      const roundCount = await prisma.question.count({ where: { issueId } });
+      if (roundCount < MAX_FOLLOWUP_ROUNDS) {
```
perrydev17, line 64: this is to guard creating endless followup questions

src/inngest/functions/reevaluate-issue-on-answer.ts
```diff
@@ -0,0 +1,71 @@
+import "server-only";
```
perrydev17, line 1: this is for reconsider/reevaluate an issue after a question is answered, without running the processInboxEmail. It runs after Inngest receive the issue/question.answered event. 

1. load the Question
2. gather the prompt for Vex again, include the new answer
3. Vex again try to classify the same issue second time.
4. get the updated data, check did the answer resolve the unknown
5. to re-triageNotification or not with VEX's fresh determination

src/features/questions/server/routers.ts
```diff
@@ -0,0 +1,112 @@
...
+  getManyByIssueId: protectedProcedure
+    .input(z.object({ issueId: z.string() }))
+    .query(({ input }) =>
+      prisma.question.findMany({
+        where: { issueId: input.issueId },
+        orderBy: { createdAt: "asc" },
+      }),
+    ),
+
+  respond: protectedProcedure
```
perrydev17, line 35: This is the api for handling user action to the Question, 3 states
1. User answer a question and submit
2. User selects not sure
3. User selects skip (Dismiss)

src/features/inbox/agent/vex/context.ts
```diff
@@ -343,6 +343,129 @@ function renderVexPrompt(args: {
   return sections.join("\n\n");
 }
 
+export async function gatherVexContextForIssue(
```
perrydev17, line 346: this serve similar purpose like gatherVexContext except it focus on one specific issue, then using the `renderAnswerContext` below to add a new section into the prompt

src/features/inbox/agent/question/index.ts
```diff
@@ -35,11 +36,31 @@ export async function generateQuestionForNotification(
...
   if (!context) return null;
   const result = await draftQuestion(context);
-
+  console.log("result ", result);
   const summary = await applyQuestionWrites(context, result);
+  console.log("summary ", summary);
+  return { ...summary, issues: context.issues.length };
+}
 
+export async function generateFollowUpQuestion(
```
perrydev17, line 48: target specific issue, generate followup question after user answered. Do we need to link the followup question to first question in someways? (like a question chain). Are we planning to show the chain questions in the UI also?

if we do, I am thinking we may need to add a parentQuestionID and a questions list or similar to the new Question model

src/features/inbox/agent/question/context.ts
```diff
@@ -201,3 +205,79 @@ export async function gatherQuestionContext(
 export const SYSTEM_PROMPT = `You are drafting clarifying questions for a hospital security engineer, for vulnerability issues a triage agent already marked "under investigation" because it lacked enough information to decide if the vulnerability is exploitable.
 For each issue, use the stated reason it's under investigation to write ONE specific, answerable question - never a vague "please provide more information." Include 2-6 short suggested answers a user could pick instead of typing. Ground every question in the evidence given; never invent facts about the device or vulnerability.
 Omit any issue you don't have a good, specific question for.`;
+
+export async function gatherQuestionContextForIssue(
```
perrydev17, line 207: similar to gatherQuestionContext does, but scoped to one already-known issue instead of whole notification. It takes extra prior Q&A

src/inngest/functions/process-inbox-email.ts
```diff
@@ -397,18 +397,20 @@ export const processInboxEmail = inngest.createFunction(
...
-    // 11. Generte questions for any Issue VEX just left UNDER_INVESTIGATION
-    const questionSummary = await step.run("generate-questions", async () => {
-      if (!notificationId || !vexSummary)
-        return { questionSkipped: true as const };
-      if ("vexSkipped" in vexSummary && vexSummary.vexSkipped)
-        return { questionSkipped: true as const };
-      return generateQuestionForNotification(notificationId);
-    });
+    // run question and mitigation steps in parallel
+    const [questionSummary, , mitigationSummary] = await Promise.all([
```
perrydev17, line 401: update from our discussion in the previous pr https://github.com/PATCH-UPGRADE/viper/pull/180/changes#r3646956200

src/features/inbox/agent/question/index.ts
0xcad: > Do we need to link the followup question to first question in someways?
> Are we planning to show the chain questions in the UI also?

Great point, I had not scoped that out yet. I think the answer to both should be yes though. 

Here's how we do something similar with Artifact
```
model Artifact {
  ...
  prevVersionId String? @unique // null for v1
  prevVersion Artifact? @relation("VersionChain", fields: [prevVersionId], references: [id], onDelete: SetNull)
  nextVersion Artifact? @relation("VersionChain") // opposite side of prevVersion

   versionNumber Int // 1, 2, 3...
  ...
}
```

we can have `parentQuestionId` / `parentQuestion` and `followupQuestion` fields. Alternatively, if you think it's better, we can create a new model `QuestionChain`, and each question in a chain has a fk to `QuestionChain` and an `order` integer?

In the UI, we can:
* fetch all questions linked to one notification
* group them in memory (either via something like `parentQuestion` + `order`, or ` QuestionChain` model)

0xcad: Happy to leave that to the next PR

src/features/inbox/agent/question/index.ts
```diff
@@ -22,7 +24,6 @@ export async function draftQuestion(
   const model = new ChatAnthropic({
     model: MODEL,
     maxTokens: 4000,
-    thinking: { type: "enabled", budget_tokens: 2000 },
```
0xcad, line 25: +1, good catch

src/features/inbox/agent/question/index.ts
0xcad, line 18: Something I just caught: This `draftQuestion` model never sees the original email that created a notification (or pdf attachments), just the `notification` (which at this point, includes only the extracted summary)

in PR 3/3, can you add the PDF attachments? I think that will give the model better information to ask questions, especially since VEX agent (which marks an issue as UNDER_INVESTIGATION in the first place) already has access to that info

src/features/inbox/agent/vex/context.ts
```diff
@@ -343,6 +343,129 @@ function renderVexPrompt(args: {
...
+
+  const cveById = new Map<string, string>();
+  cveById.set(
+    issue.vulnerability.id,
+    issue.vulnerability.cveId ?? issue.vulnerability.id,
+  );
+
+  let markdown = renderVexPrompt({
+    sources: [],
+    vulnerabilities: [issue.vulnerability],
```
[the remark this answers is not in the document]
0xcad: @perrydev17 I would address this in 3/3 (perhaps by just adding a flag to `renderVexPrompt` that you can pass *here* to make sure `_No source text._` does not get rendered

See also https://github.com/PATCH-UPGRADE/viper/pull/183/changes#r3661139584

src/inngest/functions/reevaluate-issue-on-answer.ts
```diff
@@ -0,0 +1,71 @@
...
+    await step.run("resort", async () => {
+      const context = await gatherVexContextForIssue(
+        issueId,
+        question.notificationId,
+        {
+          title: question.title,
+          reasonWhy: question.reasonWhy,
+          answer: question.answer!,
+        },
+      );
```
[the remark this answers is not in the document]
0xcad: @perrydev17

0xcad (approved):

src/features/inbox/agent/question/index.ts
perrydev17: yes I can add that in the next pr

PR #190, opened 2026-07-29
[VW-396] Seed script for testing the inbox agents against a real advisory

buggyhuy:
[VW-396](https://northeastern-patch.atlassian.net/browse/VW-396)

Adds `scripts/seed-advisory-environment.ts`, a dev script that prepares VIPER to receive one of two real Siemens Healthineers advisories by email and drive the full `processInboxEmail` chain — VEX → questions → triage → mitigation — against a hospital fleet that actually exists.

Two assets are set up to be excepted by VEX on grounds taken from the advisories themselves: `pacs-syngo-03` was hot-fixed to `VB30E_HF07` while the CMDB went stale, and `mri-magnetom-05` runs in the "workstation mode" DOC 2 describes with ports 32912/32914 closed. Everything else should stay `AFFECTED`.

Should cover all 3 cases:
- If notification seed file already seeded.
- If notification seed file haven't seeded.
- if this seed file has already seeded.

By running this script, we should aways expect to be able to test out the full flow.

## Testing

Run against a local stack (`npm run db:seed` must have run first — the script needs `user@example.com`):

```bash
npx tsx scripts/seed-advisory-environment.ts
```

The script asserts its own invariants and exits non-zero if any fail: zero notifications, one baseline `Issue` per `DeviceGroupMatching` per CVE, zero asset-level issues, and exactly two exception notes on exactly the two intended hosts.

Quick checks:

- **Clean run** — 3 syngo.plaza assets / 1 matching, 11 deserialization assets / 18 matchings; baseline issues equal matchings for both CVEs.
- **Idempotency** — three consecutive runs produce identical row counts across assets, device groups, matchings, notes, remediations, issues, and notifications.
- **Matching resolution**, using the real `matchingAppliesToDeviceGroup`: every seeded asset is reachable from its matching, including both `vers:generic/…` ranges (MAGNETOM → 3, syngo.via → 2, SOMATOM go.Top → 1).
- **Contamination** — ran `seed-notifications.ts` first to leave its note on `pacs-syngo-02`, confirmed teardown removes it and exactly the two intended notes remain. Without this, DOC 1 would return two `NOT_AFFECTED` hosts.

Results:
Screenshot 2026-07-29 at 3 53 39 PM

Screenshot 2026-07-29 at 3 51 08 PM

Screenshot 2026-07-29 at 3 51 01 PM

**Note**:
- This works for [this doc]:https://drive.google.com/file/d/1St0d8Q3w6FeYzAyxCDWb5smD5_aQCu9H/view
- Not work for [this doc]:https://drive.google.com/file/d/16d-hLOfWqm6L-sG9aMnW3oADuSu5ESBE/view
^ this due to the agent can't not extract machines or fixes out of the doc. 

[VW-396]: https://northeastern-patch.atlassian.net/browse/VW-396?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/email-pipeline-setup.md
+++ b/docs/email-pipeline-setup.md
@@ -100,23 +100,17 @@ The pipeline makes several Anthropic calls (relevance triage, classification, en
 
 ## 7. Prepare the hospital environment (optional)
 
-To exercise the VEX, triage and mitigation agents, the platform needs assets, device groups and vulnerabilities the incoming advisory can actually link to. `scripts/seed-advisory-environment.ts` sets that up for two real Siemens Healthineers advisories:
-
 ```bash
 npx tsx scripts/seed-advisory-environment.ts
 ```
 
-It **deletes every notification** and both advisory vulnerabilities, then rebuilds the substrate: device groups, assets, remediations, and notes that give the VEX agent grounds to mark some assets `NOT_AFFECTED` while their siblings stay `AFFECTED`. It never creates a notification — that has to come from the pipeline, which is the point.
-
-The script prints a summary and exits non-zero if the environment came out unusable. It requires `npm run db:seed` to have run first (it needs `user@example.com`) and refuses to run against a non-local `DATABASE_URL`.
-
-Don't run it alongside `scripts/seed-notifications.ts` — that script creates notifications for the same two advisories, which is exactly the state this one clears.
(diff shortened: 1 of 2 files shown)


perrydev17 (approved):
thanks, looks good and will give it a try later

PR #194, opened 2026-07-30
[VW-389] Note agent to create, update or delete Note when user marks affected product is incorrect

perrydev17:
ticket: https://northeastern-patch.atlassian.net/browse/VW-376

## Summary 

  * Added automated note creation, updating, and soft deletion based on MatchFeedback.
  * Added support for retaining note history without permanently removing records. (soft delete)
  * Added background processing for note actions with duplicate-request protection.
  * Fixed Affected Products details now appear only when matching products exist.
  * Deleted notes are excluded from relevant note results.
  * Note generation is triggered for non-empty match feedback comments.
  
### Create a note from unlink affected product (MatchFeedback object)

https://github.com/user-attachments/assets/7e737f7a-d8f0-43ef-838b-2c1e7711336f

### Update a note

https://github.com/user-attachments/assets/fcd910de-ba71-4f9e-bf36-97612fce98c3

### Soft Delete a note 

https://github.com/user-attachments/assets/1c68b6e4-f6e9-4fed-99c7-4ae4bd637b87

### Fixed UI when users unlink all affected product, Affected Product section is empty
Before
EmptyAffectedProducts

After

https://github.com/user-attachments/assets/6c832691-9389-4572-a8ea-a02b645d7888

## Summary by CodeRabbit

* **New Features**
  * Added automated note suggestions that can create, update, or remove notes based on relevant feedback and context.
  * Added soft deletion for notes, preserving history while removing deleted notes from relevant results.
  * Note actions now process asynchronously for reliable background updates.

* **Bug Fixes**
  * The “Affected Products” section is now hidden when no products are associated with a notification.
  * Non-empty feedback comments now trigger note processing automatically.


src/features/inbox/components/notification-details-tab.tsx
```diff
@@ -226,67 +226,71 @@ export function NotificationDetailsTab({
     }
   };
 
+  const hasAssets = withAssets.length > 0;
```
perrydev17, line 229: I ran into empty section after unlink all the assets, add a small fix to not show this section

Image

src/features/notes/agent/noteAction/context.ts
```diff
@@ -0,0 +1,90 @@
...
+  const markdown = renderNoteActionPrompt({ request, candidates });
+
+  return { request, markdown, candidates };
+}
+
+export const SYSTEM_PROMPT = `You maintain a hospital vulnerability-management team's notes about their medice devices, assets, vulnerabilities, and remediations. Notes are the team's durable memory: they can fed back into every later AI run and shown to staff on the asset, vulnerability, and remediation pages. A user has just taken an action and left a comment.
+
+Your job is to record what they told us, do not judge whether a comment is "important enough" and emit note operations. Any claim about a device, vulnerability, or remediation must come from "what the user said". Never invent a technical or clinical claim the user did not make.
+
+What to record:
```
perrydev17, line 75: Should we just record everything, given that Note is intended to store information coming from different sources?

VW-393 can take some of these prompt/handlings potentially

src/features/notes/agent/noteAction/process_output.ts
```diff
@@ -0,0 +1,110 @@
...
+      claimed.add(write.noteId);
+    }
+
+    writes.push(write);
+  }
+
+  return writes;
+}
+
+export async function applyNoteAction(
```
perrydev17, line 65: I think, in a separate ticket, we can create createNote, updateNote, deleteNote helper functions and use them across different areas in the codebase where there are note actions.

src/features/notes/agent/noteAction/context.ts
0xcad: These instructions are somewhat self-contradictory, with

> Your job is to record what they told us, do not judge whether a comment is "important enough" and emit note operations

and

> What to record:

which implies that some things are more important.

I don't think we need to record everything -- I think the agent should have a choice to omit notes. As the best decision would come from seeing user behavior though, I think for now we can leave a TODO comment representing this consideration and just move on with the PR

VW-393 will involve the chat agent using a tool to record notes (or trigger a subagent to record notes for it) on seeing only key info that should be saved. so VW-393 an agent "omits a note" by just not using the manage_notes tool

src/features/notes/agent/noteAction/process_output.ts
```diff
@@ -0,0 +1,110 @@
...
+  candidateIds: Set<string>,
+): NoteActionWrite | null {
+  const text = (op.text ?? "").trim();
+
+  if (op.action === "create") {
+    return text ? { action: "create", text } : null;
+  }
+
+  const noteId = op.noteId;
+  if (!noteId || !candidateIds.has(noteId)) return null;
```
0xcad: consider, instead of returning null when model has incorrect output (either no noteId, or an incorrect noteId), throw an error (which would trigger inngest to retry the function) or at least console.warn for observability

src/features/notes/agent/noteAction/context.ts
0xcad: For VW-393, you can create a tool that asynchronously calls this agent (using the inngest event + `resolveNoteActionRequest`), the input can be the user's comment + perhaps a relevant `targetModel` + `instanceId`. Then this is the "subagent" that handles deduplication/update/deleting notes

So again:
* chat agent has a tool that triggers an inngest event, supplies information about the user's comment and what it should be linked to
* this agent looks through the catalogue of notes and determines if something should be updated/deleted, or to just create a new note

0xcad (approved):

src/features/notes/agent/noteAction/process_output.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  return writes;
+}
+
+export async function applyNoteAction(
+  context: NoteActionContext,
+  result: NoteActionResult,
+): Promise<NoteActionSummary> {
+  const writes = planNoteActions(context, result);
+  console.log("writes, ", writes);
```
buggyhuy: minor, it seems like for debugging purpose, can you remove it?

perrydev17: thanks, removed

src/features/notes/agent/noteAction/process_output.ts
perrydev17: updated

src/features/notes/agent/noteAction/context.ts
perrydev17: added a TODO, and will update this in VW-393

PR #200, opened 2026-08-03
[VW-390] Plain-language agent output and an evidence-based triage tier procedure

buggyhuy:
[VW-390](https://northeastern-patch.atlassian.net/browse/VW-390)

## Summary
- Pulling Internal Ids out to a Ref Module.
- Now only Mitigation Plans Builder Agent has access to Ref Module to get Linking Ids.
- Added Guard for Failure of converting ref > actual ids. This will trigger a re-try + error so we can monitor

## Testing

**End-to-end (live stack, real Anthropic + real Resend):** advisory Siemens SSA-016040 / `CVE-2024-52334` through `processInboxEmail` on the VW-396 seed environment. VEX created 1 asset-level override (`pacs-syngo-03` → confirmed unaffected, `COMPONENT_NOT_PRESENT`); mitigation produced 2 plans / 5 draft work orders with 0 dropped links. Zero database ids and zero status-code strings in any persisted prose field.

Updated: 
Screenshot 2026-08-04 at 4 29 10 PM
Screenshot 2026-08-04 at 4 29 20 PM
Screenshot 2026-08-04 at 4 30 59 PM
Screenshot 2026-08-04 at 4 34 58 PM
Screenshot 2026-08-04 at 4 35 12 PM

[VW-390]: https://northeastern-patch.atlassian.net/browse/VW-390?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Added clearer, hospital-staff-friendly guidance for triage, mitigation, and device-impact explanations.
  * Improved work-order and device references with readable, structured links.
  * Enhanced device-group, asset, and vulnerability context in issue summaries.

* **Bug Fixes**
  * Prevented internal IDs, platform terminology, and raw score lists from appearing in staff-facing responses.
  * Improved priority and impact explanations by highlighting verification status and deciding factors.

* **Tests**
  * Expanded validation for references, link conversion, readable plan content, and invalid identifiers.


src/features/inbox/agent/triage/index.ts
```diff
@@ -21,24 +21,40 @@ export type TriageResult = z.infer<typeof triageSchema>;
 
 const SYSTEM_PROMPT = `You are a triage agent for a hospital cybersecurity platform. Given a security notification and the resolved hospital context, assign a priority tier, explain why, and describe the clinical and operational impact on the hospital.
 
-PRIORITY TIERS:
+PRIORITY — decide the tier with this procedure, in order. It runs on the evidence alone: how you word the text fields must never change its outcome.
```
buggyhuy: Grading so that the result is consistent. The grading rubric is based on [enrich-vulnerabilities.ts#L14](https://github.com/PATCH-UPGRADE/viper/blob/c0ba4b292b540f9f18b1f4e2c8938a9a75acef19/src/inngest/functions/enrich-vulnerabilities.ts#L14)

src/features/inbox/agent/mitigation/index.ts
```diff
@@ -23,14 +23,19 @@ const SYSTEM_PROMPT = `You are a mitigation-planning agent for a hospital cybers
 
 Each plan is a coherent strategy for reducing the risk from this notification, made up of concrete work orders.
 
+AUDIENCE & VOICE — plan titles, summaries, compareLines, cards, and work-order descriptions are shown directly to hospital staff (administrators, clinicians, biomedical engineers), NOT to security engineers:
+- Never put database ids in any prose field. Ids belong ONLY in the structured linking fields (vulnerabilityIds, remediationIds, deviceGroups). In prose, name things the way staff know them: the CVE number, the vendor and product (e.g. "the syngo.plaza VB30 workstations"), the remediation by its title, the location.
+- Never use internal vocabulary in prose: no "VEX", no statuses like "NOT_AFFECTED" or "UNDER_INVESTIGATION", no "device group matching". Say "confirmed unaffected", "still being verified", "these devices".
```
0xcad: We didn't want the model to be be referencing internal ID's in its outputs to users, nor referring to "VEX", which is the name of an internal model (inspired by [Vulnerability Exploitability eXchange](https://cyclonedx.org/capabilities/vex/)) that could be potentially be confusing to users.

The model needs to have information on internal ID's so it can link assets/vulnerabilities/remediations to tickets that it creates

Your solution is just modifying the system prompt to tell the model not to use these in prose fields

An alternate solution worth testing is to break up the prompt into two pieces:
1. Model that writes tickets, and cannot see internal ID's
2. Model that links tickets to assets/vulns/remediations, and can see internal ID's

This way, the model _cannot_ write assetId's into internal tickets, because it doesn't have those tickets in the first place.

By the way, if you can think of a better way to break up the model prompts instead of my #1 or #2 above, feel free to try that instead

With "VEX" issues, another approach is modifying the prompt so that:
1. We no longer call them "VEX"
2. We wrap them in something like an `<issues></issues>` tag, and tell the LLM that information inside of `<issues></issues/>` is only for it to triage vulnerabilities, and should not be referenced in created tickets

src/features/inbox/agent/mitigation/index.ts
```diff
@@ -23,14 +23,19 @@ const SYSTEM_PROMPT = `You are a mitigation-planning agent for a hospital cybers
...
+- Lead with the action and its consequence for the hospital, in plain words; keep technical detail as brief supporting explanation. A work-order detailedDescription should read like an instruction a colleague could hand to the biomedical engineering team.
```
0xcad, line 29: I like this

src/features/inbox/agent/mitigation/index.ts
```diff
@@ -23,14 +23,19 @@ const SYSTEM_PROMPT = `You are a mitigation-planning agent for a hospital cybers
...
+
 RULES:
 - Order the plans best-first: the FIRST plan is the one you recommend. Typically 1-3 plans.
 - Plans must be genuinely different strategies (e.g. contain-now-via-network vs. patch-every-device), not slight variations. Give each a "compareLine" that says how it trades off against the others.
 - Ground EVERY field in the provided notification and hospital context. Never invent device counts, CVSS/EPSS numbers, care areas, downtime figures, or remediations — use only what the context states. If the context is thin, keep the cards qualitative rather than fabricating numbers, or report "Uncertain"
-- Prefer plans that map to the linked remediations and affected device groups. Factor VEX determinations in: assets marked NOT_AFFECTED do not need remediating.
+- Prefer plans that map to the linked remediations and affected device groups. Factor VEX determinations in: assets marked NOT_AFFECTED do not need remediating (in prose, call them "confirmed unaffected").
```
0xcad: again, I would just change "NOT_AFFECTED" in the actual prompt to the LLM as "confirmed unaffected". Don't show the LLM "NOT_AFFECTED" if you don't want it to say NOT_AFFECTED, show it what we want it to say instead

src/features/inbox/agent/mitigation/index.ts
```diff
@@ -23,14 +23,19 @@ const SYSTEM_PROMPT = `You are a mitigation-planning agent for a hospital cybers
...
+- Never put database ids in any prose field. Ids belong ONLY in the structured linking fields (vulnerabilityIds, remediationIds, deviceGroups). In prose, name things the way staff know them: the CVE number, the vendor and product (e.g. "the syngo.plaza VB30 workstations"), the remediation by its title, the location.
+- Never use internal vocabulary in prose: no "VEX", no statuses like "NOT_AFFECTED" or "UNDER_INVESTIGATION", no "device group matching". Say "confirmed unaffected", "still being verified", "these devices".
+- Lead with the action and its consequence for the hospital, in plain words; keep technical detail as brief supporting explanation. A work-order detailedDescription should read like an instruction a colleague could hand to the biomedical engineering team.
+
 RULES:
 - Order the plans best-first: the FIRST plan is the one you recommend. Typically 1-3 plans.
 - Plans must be genuinely different strategies (e.g. contain-now-via-network vs. patch-every-device), not slight variations. Give each a "compareLine" that says how it trades off against the others.
```
0xcad, line 33: Alternate idea worth trying to see how it performs again: one prompt call creates tickets, the other prompt call links tickets via ID's. Model that creates tickets cannot see asset ID's

Or maybe:
* First prompt creates tickets and links them via ID's, but does not write in prose fields. Instead, it writes a plan for the second LLM
* Second LLM gets the plan, and work orders, then fills in the work order prose fields with instructions

src/features/inbox/agent/triage/index.ts
```diff
@@ -21,24 +21,40 @@ export type TriageResult = z.infer<typeof triageSchema>;
...
+
+Step 1 — Start from the exploitation evidence:
+- On CISA KEV and CVSS >= 7.0 -> Critical
+- On CISA KEV and CVSS < 7.0 -> Monitor
+- Not on KEV, EPSS >= 0.088 (8.8%), CVSS >= 7.0 -> High
+- Not on KEV, EPSS >= 0.088 (8.8%), CVSS < 7.0 -> Monitor
+- Not on KEV, EPSS < 0.088 (8.8%) -> Defer
+- If EPSS or CVSS is absent, pick the closest row from the strongest evidence the context does state (public exploit code, reports of active exploitation, vendor severity rating).
```
0xcad: FYI, never have an LLM "do math" like this. If you don't need an LLM to solve a problem, do not use an LLM. You do not need an LLM to apply a priority with a deterministic algorithm (`enrich-vulnerabilities.ts` already does this without an LLM)

I agree that priorities are not consistent and need work. That's not in the scope of this ticket though

For reference/learning though, if you *did* have to change the priority in this ticket, you would:
* Re-use the decision tree logic in enrich-vulnerabilities, with KEV/EPSS/CVSS to set the priority of a notification without an LLM
* If EPSS/CVSS was absent, in this case you could use an LLM to "pick the closest row from the strongest evidence"

0xcad (changes requested):
1. Priority prompt changes are out of scope, this ticket just aims to make LLM's use more plain language via prompting, and stop referencing internal ID's/db structure, again, through prompting and changing prompt structures
2. Look at comments on line 28/35, which are relevant to other parts of this code, too. How can you get it to avoid using internal ID's/internal language, by either:
    * Changing the name of that internal language into something prose friendly
    * Making internal ID's not visible to the agent writing work orders, by breaking down prompts further into multiple steps
    * Consider using something like an `<issues></issues>` wrapper around the VEX determinations and describing rules about how the model should interact with text wrapped by that

src/features/inbox/agent/triage/index.ts
buggyhuy: - I agree. It could very likely hallucinate on the math. 
- I was lumping this in while trying to test the feature, but totally understand why this is out of scope here.
- I will roll this back, and we can address this in  the future.

src/features/inbox/agent/mitigation/index.ts
buggyhuy: Noted it, good advice.

src/features/inbox/agent/mitigation/index.ts
buggyhuy: Sounds good, let me try to break them up.

src/features/inbox/agent/mitigation/index.ts
buggyhuy: good idea. Let me try that out.

buggyhuy:
Proposed a new solution: Using a reference module. This way Internal Ids won't make it into the prompt, unless we intentionally provide context to swap it out for real value (which happens in Mitigation Plan). 
- Added a guard so we don't accidentally failed the conversion.

src/features/inbox/agent/mitigation/index.ts
```diff
@@ -19,24 +19,24 @@ import {
...
-- LINKING: every work order names the vulnerabilities, remediations, and device groups IT specifically addresses, using the exact ids shown inline in the hospital context (e.g. "(id: …)" on a vulnerability heading, the id on a remediation heading, "(id: …)" on a device group line). Never invent an id. These ids go ONLY in the linking fields — never repeat them in shortDescription, detailedDescription, or any other text.
-- Link narrowly. A work order that firewalls one device group must NOT list the other groups, and one that applies a single vendor patch must NOT list every remediation. Different work orders in a plan usually target different device groups. Leave a list empty rather than padding it.
+- LINKING: every work order names the vulnerabilities, remediations, and device groups IT specifically addresses, using the exact reference codes shown inline in the hospital context (e.g. "(id: vuln-1)" on a vulnerability heading, "rem-1" on a remediation heading, "(id: group-2)" on a device group line). Never invent a reference. These codes go ONLY in the linking fields — the prose names things by CVE, manufacturer/vendor, product, and location.
```
0xcad, line 38: I like this

src/features/inbox/agent/mitigation/index.ts
```diff
@@ -124,8 +124,16 @@ export async function createMitigationPlans(
     );
   }
 
-  return {
-    plans: parsed.data.plans as MitigationPlanItem[],
-    linkableIds,
-  };
+  const toId = (ref: string) => refs.idByRef[ref];
```
0xcad, line 127: Error handling? What happens if `ref` is hallucinated?

src/features/inbox/agent/mitigation/schema.test.ts
```diff
@@ -98,24 +102,70 @@ describe("buildMitigationPlansSchema", () => {
...
+  it("does not flag prose that merely resembles a ref", () => {
+    const parsed = schema.safeParse(
+      withWorkOrder({
+        ...goldenPlan.workOrders[0],
+        detailedDescription:
+          "Move the scanners to VLAN group-10 and verify subgroup-1 connectivity.",
+      }),
+    );
+    expect(parsed.success).toBe(true);
+  });
```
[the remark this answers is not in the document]
0xcad: I modified CLAUDE.md to avoid getting these coderabbit suggestions, hopefully they will stop soon

0xcad (approved):

src/features/inbox/agent/mitigation/index.ts
buggyhuy: A hallucinated ref, if I understand it as a made up ref like "vuln-8" but we don't have an "vuln-8"; shouldn't be able to get here.

gatherTriageContext pulls data from DB, setup all these "vuln-1", "rem-1", etc thru buildEntityRefs(linkableIds) - which then put them as z.enum([...]). Then pass them to buildMitigationPlansSchema types them as z.enum([...]) over exactly the refs minted for this request.

so a made-up vuln-8 (or a raw database id) fails the schema.safeParse a few lines up and we throw there, which fails the Inngest step and re-rolls the model.

PR #206, opened 2026-08-04
[VW-405] Asset detail page has button to get a PDF with a QR code on it

trunggo9000:
ticket: https://northeastern-patch.atlassian.net/browse/VW-405

## Summary

- Adds a "QR Code Printout" link to the asset detail page header, the asset dashboard drawer header, and a new "Links" section in the drawer's Details tab

## Screenshts
Asset QR Code Printout pdf
image

Asset detail page with QR Code printout header
image

Asset dashboard with QR Code printout header
image

## Summary by CodeRabbit

* **New Features**
  * Added the ability to generate and open printable QR-code PDFs for assets.
  * QR PDFs include equipment details, branding, and optional serial numbers.
  * Added QR/PDF access from both the asset header and details panel.

* **Bug Fixes**
  * Improved work-order table readability by wrapping long summaries and aligning ticket details consistently.

PR #208, opened 2026-08-05
[VW-405] fix QR Code pdf generation

trunggo9000:
ticket: https://northeastern-patch.atlassian.net/browse/VW-405

## Summary

- The QR PDF feature (added in #206) rendered via headless Chromium  which turned out to be too much memory for a Vercel function and was failing to generate the PDF in production
- Replaces it with `pdf-lib`: draws the same layout directlwith no browser launch involved 

## Screenshot
image

## Summary by CodeRabbit

* **New Features**
  * QR code PDFs are now generated directly with improved reliability and without requiring browser-based rendering.
  * Generated PDFs include QR codes, standard document content, optional serial data, and branding.

* **Bug Fixes**
  * Improved browser startup handling for fleet session capture across local and hosted environments.

* **Tests**
  * Updated PDF rendering tests to run explicitly in a Node environment.


src/features/assets/server/qr-pdf.ts
```diff
@@ -20,137 +25,134 @@ export const qrPdfAssetSelect = {
 
 type QrPdfAsset = Prisma.AssetGetPayload<{ select: typeof qrPdfAssetSelect }>;
 
-const logoSvg = fs.readFileSync("public/logos/logo.svg", "utf-8");
+const PAGE_WIDTH = 612;
```
perrydev17, line 28: maybe it is just on my screen, the link text was appears to be cut in the pdf. Maybe consider using a dynamic width instead of a fixed one?
https://pdf-lib.js.org/docs/api/interfaces/pdfpagedrawtextoptions

Image

0xcad: same issue. @trunggo9000 if you can fix it, great, if you can't, at least the real environment (`viper-xi.vercel.app`) is a shorter URL that won't overflow

0xcad (approved):
see Perry's comment about using a dynamic width instead of a fixed one for drawing the URL text, which currently gets cut off the screen in vercel preview environments

PR #209, opened 2026-08-05
[VW-393] Create record_note tool for chat agent, allows the system to create/update/delete a note with user chat input 

perrydev17:
ticket: https://northeastern-patch.atlassian.net/browse/VW-393

## Summary

  * Added a new tool for chat agent to record relevant info provided by user.
  * Added the ability to create/update/delete notes through chat info provided by user.
  * Supports notes for specific records and broader filtered scopes.
  * Automatically identifies note targets and displays clear labels.
  * Added guidance to prevent duplicate notes and capture durable facts appropriately.
  * Notes are recorded asynchronously with confirmation messaging.

### Out of scope
- EntityFilter creation

https://github.com/user-attachments/assets/fb3a9fa4-4b56-43c9-be60-9a329d2d1089

## Summary by CodeRabbit

- **New Features**
  - Added chat support for recording persistent and target-specific notes.
  - Notes can be linked to assets, vulnerabilities, remediations, or device-group matches with clear target labels.
  - Added validation, duplicate avoidance, and asynchronous note processing.
  - Improved note context for AI-generated actions with durable facts and appropriate scoping.

- **Bug Fixes**
  - Improved handling of invalid note content and unresolved targets.
  - Added clearer target labels to inbox question context.

### DIFF
--- a/src/features/chat/viper-agent/langgraph/chat-graph.ts
+++ b/src/features/chat/viper-agent/langgraph/chat-graph.ts
@@ -66,7 +66,7 @@ Scope it. A fact about one device goes on that asset. A fact covering every devi
 goes on the DEVICE_GROUP_MATCHING, not on one asset.
 
 You do not choose create vs update vs delete, the notes agent does, after reading what already exists.
-Recording is asynchronous, so never tell the user a note was created. Always state in your replay what you recorded,
+Recording is asynchronous, so never tell the user a note was created. Always state in your reply what you recorded,
 in one short sentence (e.g. "I've noted that these ventilators run firmware 3.2").
 The sentence is what carries the fact forward in this conversation.
 `;
--- a/src/features/chat/viper-agent/langgraph/chat-graph.ts
+++ b/src/features/chat/viper-agent/langgraph/chat-graph.ts
@@ -26,6 +26,7 @@ Be concise, accurate, and prioritize patient safety in your recommendations.
 - list_fleet_managed_assets: list the assets Siemens Healthineers services.
 - propose_fleet_work_order: propose a work order on Siemens Healthineers'
   teamplay Fleet platform. Your turn ends here until the user accepts or dismisses.
+- record_note: record a durable fact the user tells you about. Fire and forget, a separate notes agent decides whether it creates, updates or deletes a note.
 </tools>
 
 ## Data access
@@ -49,8 +50,26 @@ tell the user to phone Siemens — those can't be filed online.
 
 ## Notes
 Persistent hospital-wide notes are provided below as context. Treat them as authoritative
-standing facts about this hospital.`;
-// TODO: VW-393: Add new tool for note management
+standing facts about this hospital. Entity-specific notes come back on the records you fetch 
+with query_platform_data, in each record's "notes" array.
+
+## Recording notes
+When the user tells you something durable about their fleet, record it with record_note: 
+fleet composition, configuration or exposure specifics, standing operational constraints,
+or a correction to how a device or vulnerability should be read going forward.
+Do NOT record one-time questions, what you just looked up, or your own analysis.
+
+Check first. If the fact is already in the record's "notes" array or in the hospital-wide notes below,
+say so instead of recording it again. One atomic fact per call, split a message carrying two unrelated facts into two calls.
+
+Scope it. A fact about one device goes on that asset. A fact covering every device of a make/model
+goes on the DEVICE_GROUP_MATCHING, not on one asset.
+
+You do not choose create vs update vs delete, the notes agent does, after reading what already exists.
+Recording is asynchronous, so never tell the user a note was created. Always state in your replay what you recorded,
+in one short sentence (e.g. "I've noted that these ventilators run firmware 3.2").
+The sentence is what carries the fact forward in this conversation.
+`;
 
 function buildSystemPrompt(role: UserRole): string {
   return [
--- a/src/features/chat/viper-agent/langgraph/note-tool.ts
+++ b/src/features/chat/viper-agent/langgraph/note-tool.ts
@@ -6,7 +6,6 @@ import { ChatNoteInput } from "@/features/notes/types";
 import { ScopeTargetModel } from "@/generated/prisma";
 import { requestNoteAction } from "@/inngest/functions/notes-action";
 
-// https://code.claude.com/docs/en/agent-sdk/custom-tools
 export function makeRecordNoteTool(userId: string) {
   return tool(
     async ({ text, targetModel, instanceId }) => {
@@ -31,7 +30,7 @@ export function makeRecordNoteTool(userId: string) {
         input,
       );
       if (!queued) {
-        return `NOT RECORDED. Tell the user their note did not save and to try again in a momment. Do you claim you record anything.`;
+        return `Failed to queue. Nothing was recorded.`;
       }
 
       return [
@@ -42,12 +41,7 @@ export function makeRecordNoteTool(userId: string) {
     },
     {
       name: "record_note",
-      description: `Record a durable fact the user told you about, so it is available to staff and to every later AI run.
-            Use this to record, correct, or retract information. A separate notes agent reads the notes that already exist on the target and decides whether to create a new note, update an existing one, or delete one.
-            Record: fleet composition, configuration or exposure specifics, standing operational constraints, and corrections to how a device or vulnerability should be read going forward.
-            Do NOT record transient chat: one off questions, what you just looked up, or your own analysis.
-            Before calling, check the "notes" array already returned on that record by query_platform_data - if the fact is there, say so instead of recording it again.
-            One atomic fact per call. Recording is asynchronous, so never tell the user a specific note was created, say you have recorded the fact.`,
+      description: `Record a durable fact the user told you about, so it is available to staff and to every later AI run. Use this to record, correct, or retract information. A separate notes agent reads the notes that already exist on the target and decides whether to create a new note, update an existing one, or delete one. Before calling, check the "notes" array already returned on that record by query_platform_data - if the fact is there, say so instead of recording it again.One atomic fact per call. Recording is asynchronous, so never tell the user a specific note was created, say you have recorded the fact.`,
       schema: z.object({
         text: z
           .string()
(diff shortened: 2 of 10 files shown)


src/features/chat/viper-agent/langgraph/note-tool.ts
```diff
@@ -0,0 +1,70 @@
...
+      const input: ChatNoteInput = {
+        comment: statement,
+        targetModel,
+        instanceId,
+        userId,
+      };
+
+      const queued = await requestNoteAction(
+        "CHAT",
+        crypto.randomUUID(),
```
perrydev17, line 29: only for idempotency purposes

src/features/chat/viper-agent/langgraph/note-tool.ts
```diff
@@ -0,0 +1,70 @@
...
+    async ({ text, targetModel, instanceId }) => {
+      const statement = text.trim();
+      if (!statement) return "Text was empty";
+
+      const targetLabel = await resolveNoteTargetLabel(targetModel, instanceId);
+      if (!targetLabel) {
+        return `No ${targetModel} exists with id ${instanceId}`;
+      }
+
+      const input: ChatNoteInput = {
```
perrydev17, line 20: worth to add threadID?

src/features/chat/viper-agent/langgraph/note-tool.ts
```diff
@@ -0,0 +1,70 @@
...
+        userId,
+      };
+
+      const queued = await requestNoteAction(
+        "CHAT",
+        crypto.randomUUID(),
+        input,
+      );
+      if (!queued) {
+        return `NOT RECORDED. Tell the user their note did not save and to try again in a momment. Do you claim you record anything.`;
```
0xcad: > momment. Do you claim you record anything

typo. i believe should read, "moment. Do not claim you recorded anything"

src/features/chat/viper-agent/langgraph/note-tool.ts
```diff
@@ -0,0 +1,70 @@
+import "server-only";
+import { tool } from "@langchain/core/tools";
+import { z } from "zod";
+import { resolveNoteTargetLabel } from "@/features/notes/server/note-targets";
+import { ChatNoteInput } from "@/features/notes/types";
+import { ScopeTargetModel } from "@/generated/prisma";
+import { requestNoteAction } from "@/inngest/functions/notes-action";
+
+// https://code.claude.com/docs/en/agent-sdk/custom-tools
```
0xcad: pretty sure there docs are not relevant (tool for anthropic-ai/claude-agent-sdk, but we use langchain/core/tools). am I missing something?

src/features/chat/viper-agent/langgraph/note-tool.ts
0xcad: I don't think it's necessary, we don't (maybe, yet) attach notes to the conversations that created them. we could consider that for the future

src/features/chat/viper-agent/langgraph/note-tool.ts
0xcad: Consider changing to "Failed to queue. Nothing was recorded" and let system prompt handle how the model should phrase that

src/features/chat/viper-agent/langgraph/note-tool.ts
```diff
@@ -0,0 +1,70 @@
...
+
+      return [
+        `Queued for ${targetModel}.`,
+        "The notes agent will compare this against the notes already on that target and may update an existing note instead of creating a new one.",
+        "Do not call record_note again for this same fact. State in your reply what you recorded, so it stays on the record for later turn.",
+      ].join(" ");
+    },
+    {
+      name: "record_note",
+      description: `Record a durable fact the user told you about, so it is available to staff and to every later AI run.
```
0xcad: This is pretty much the same as the `chat-graph` system prompt

Chat graph system prompt
* "one atomic fact per call, split a message carrying two unrelated facts into two calls" (description says "one atomic fact per call")

Recommendation: get rid of Record / Do NOT record section, which is handled in system prompt, but do keep "Use this to record, correct, or retract information. A separate notes agent reads the notes that already exist on the target and decides whether to create a new note, update an existing one, or delete one."

Also minor note, each line has indentation since it's in a ``, that means the model also sees the indentation

src/features/notes/agent/noteAction/context.ts
```diff
@@ -68,25 +67,57 @@ export async function gatherNoteActionContext(
...
+## Avoiding duplicates
+
+Two notes stating the same fact about the same thing is always a defect: both are fed into later AI runs and shown to staff, and
+nothing reconciles them. A duplicate costs more than a missed nuance, so when create and update are both defensible, update.
+
+Work through this in order before emitting any create:
+1. Does an existing note already state this fact, even in different words? => Emit nothing. Do not restate it more precisely, and do not add a second note "for clarity".
+2. Does an existing note state the same attribute with a different or older value? => update that note. This is the usual shape of a correction.
+3. Is a candidate marked "LIKELY THE SAME FACT"? => it is an update or nothing. Never a create.
+4. Does the comment repeat something under "Standing hospital-wide facts"? => Emit nothing. Those already apply everywherel a scoped copy adds noise and will drift out of sync.
```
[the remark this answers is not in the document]
0xcad: I think you do want `prisma.note.findMany({ where: { status: "PERSISTENT", deletedAt: null } })` to be included in the noteAction prompt

src/features/notes/agent/noteAction/context.ts
```diff
@@ -66,27 +65,57 @@ export async function gatherNoteActionContext(
...
+5. Only if none of the above apply => create.
+
+A comment carrying two genuinely unrelated facts becomes two operations. A comment stating one fact in two ways in one operation.
+
+## Writing the text
+
+- Name the subject. A note is stored on its own and is often read detached from the record it hangs off: e.g. "MRI-01 runs firmware 3.2" is usable, "runs firmware 3.2" is not. Take the subject from "What this note is about", and never use a pronoun for the device.
```
0xcad: Just a consideration: Consider the case where the hostname / name a user calls a device changes. In that case, *not* including the name, which can become outdated info, actually seems preferred to me. "runs firmware 3.2" does seem like a reasonable note to me.

> A note is stored on its own and is often read detached from the record it hangs off

It might be easier to just change the above behavior. `question/context.ts:56` was the only place I could find where notes are included without being attached to a label about what device it's attached to: https://github.com/PATCH-UPGRADE/viper/blob/fcaa387a69f19f4182db0218ab22896d72c6448a/src/features/inbox/agent/question/context.ts#L56

Everywhere else (that I could see) uses `renderNoteTarget(n, labels)`. `renderNoteTarget` is a better way to add labels to notes because if `MRI-01` changes in the db, `labels` just gets updated to whatever the new hostname/id is

src/features/notes/server/note-targets.ts
```diff
@@ -0,0 +1,47 @@
...
+        where: { id: instanceId },
+        select: { description: true },
+      });
+      if (!remediation) return null;
+      const firstSentence = remediation.description?.trim().split("\n")[0];
+      return firstSentence ? firstSentence : `Remediation ${instanceId}`;
+    }
+    case "DEVICE_GROUP_MATCHING": {
+      const matching = await prisma.deviceGroupMatching.findUnique({
+        where: { id: instanceId },
```
0xcad, line 36: this isn't a problem with this PR, so much as it's a problem with the code *around* it -- but I don't know where we include `DEVICE_GROUP_MATCHING` `id`s

if the chat agent would never see the device group matching id,  it can never create a note linking to a device group matching

non-applicable to this PR, but we should brainstorm how to get that to the model. perhaps we don't match based on device group matching `id`, and instead give the model, say, a json field that it can use to find-or-create a relevant device group matching

what do you think? If you agree this is a problem, can you open a ticket for it on Jira? the body can just be a link to this github comment.

src/features/chat/viper-agent/langgraph/chat-graph.ts
```diff
@@ -49,8 +50,26 @@ tell the user to phone Siemens — those can't be filed online.
...
+Do NOT record one-time questions, what you just looked up, or your own analysis.
+
+Check first. If the fact is already in the record's "notes" array or in the hospital-wide notes below,
+say so instead of recording it again. One atomic fact per call, split a message carrying two unrelated facts into two calls.
+
+Scope it. A fact about one device goes on that asset. A fact covering every device of a make/model
+goes on the DEVICE_GROUP_MATCHING, not on one asset.
+
+You do not choose create vs update vs delete, the notes agent does, after reading what already exists.
+Recording is asynchronous, so never tell the user a note was created. Always state in your replay what you recorded,
```
0xcad: typo: replay -> reply

0xcad (approved):
mostly minor comments that you can quickly address before merging, plus a suggestion to create a new ticket to address a potentially larger issue

src/features/chat/viper-agent/langgraph/note-tool.ts
perrydev17: yes, i was mainly referring to // Define a tool: name, description, input schema, handler and was looking at both langchain/claude documents. It does creates confusion, removed

src/features/notes/agent/noteAction/context.ts
perrydev17: Good point. During my initial testing, I was getting notes like "runs firmware 3.2", which didn't feel like meaningful notes, so I adjusted the prompt to avoid generating entries like that.

The example you shared is a scenario I hadn't considered, and it highlights a valid use case. I think it makes sense to account for cases like that rather than optimizing solely based on the examples I happened to encounter during testing.

fixed the prompt and also updated `question/context.ts`

src/features/chat/viper-agent/langgraph/chat-graph.ts
perrydev17: fixed

src/features/chat/viper-agent/langgraph/note-tool.ts
perrydev17: fixed

src/features/chat/viper-agent/langgraph/note-tool.ts
perrydev17: fixed

PR #211, opened 2026-08-07
[VW-419] Replace WorkOrderTicket.assets many-to-many with per-asset AssetTicket children

trunggo9000:
https://northeastern-patch.atlassian.net/browse/VW-419

## Summary

Replaces `WorkOrderTicket.assets` (a direct `Asset[]` many-to-many) with a new `AssetTicket` join model 
- New `AssetTicket` join model (`asset`, parent ticket, dedicated child ticket, cascading delete).
- Child ticket marked Done → parent marks Done once all its children are, recursing up through nested tickets.
- Deleting an asset also deletes its dedicated child ticket.

## Screenshots
image

image

image

## Summary by CodeRabbit

* **New Features**
  * Assets linked to work-order tickets now receive dedicated child tracking tickets.
  * Linked asset views display both asset details and child-ticket status, with direct navigation to ticket details.
  * Matching assets can be attached automatically during mitigation and Fleet workflows.
  * Completing linked tickets can automatically update eligible parent tickets.

* **Bug Fixes**
  * Improved duplicate prevention, hierarchy validation, status cascading, and cleanup when assets are removed.
  * Refined attachment and detachment behavior for linked assets.


prisma/schema.prisma
```diff
@@ -1217,6 +1219,26 @@ model WorkOrderTicket {
   @@map("work_order_ticket")
 }
 
+model AssetTicket {
```
buggyhuy, line 1222: Just curious for my understand. I see parentTicketId and ticketId. Is this because we support sub ticket? like there are Tickets under a parent Ticket in a scenario?

prisma/seed.ts
```diff
@@ -1,3 +1,5 @@
+import Module from "node:module";
```
perrydev17: instead of importing the app logic, consider using helper function and/or following the `seedABC` pattern ?

src/features/inbox/agent/question/context.ts
```diff
@@ -63,7 +63,7 @@ function renderQuestionPrompt(args: {
         args.notes
           .map((n) => {
             const target = renderNoteTarget(n, args.labels);
-            return target ? `- **${target}** ${n.text}`: `- ${n.text}`; 
+            return target ? `- **${target}** ${n.text}` : `- ${n.text}`;
```
perrydev17, line 66: thanks, I also have this fix in my pr

src/features/notes/agent/noteAction/context.ts
```diff
@@ -47,9 +47,8 @@ function renderNoteActionPrompt(args: {
   const sections: string[] = [];
 
   if (persistent.length > 0) {
-    sections.push(
-      `## Standing hospital wide facts\n\n `,
-    ) + persistent.map((perstn) => `- ${perstn}`).join("\n");
+    sections.push(`## Standing hospital wide facts\n\n `) +
+      persistent.map((perstn) => `- ${perstn}`).join("\n");
```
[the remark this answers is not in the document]
perrydev17: thanks, I also have this fix in my pr

src/features/tracking/server/routers.ts
```diff
@@ -87,15 +84,44 @@ function fleetContactFor(user: {
 // on `code` rather than `instanceof`: across Next.js module boundaries the thrown
 // error can be a different copy of PrismaClientKnownRequestError, so `instanceof`
 // is unreliable.
-function isUniqueViolation(error: unknown): boolean {
+function hasPrismaCode(error: unknown, code: string): boolean {
```
perrydev17, line 87: side note: we have the exact same function in `router-utils.ts` 
https://github.com/PATCH-UPGRADE/viper/blob/main/src/lib/router-utils.ts#L100

no need to change anything, this is good

src/features/mitigation/server/routers.ts
```diff
@@ -128,10 +129,27 @@ export const mitigationRouter = createTRPCRouter({
...
+            id: true,
+            deviceGroups: { select: { deviceGroupMatchingId: true } },
+          },
+        });
         await tx.workOrderTicket.updateMany({
           where: { mitigationPlanId: plan.id },
           data: { isDraft: false },
         });
+        for (const ticket of promoted) {
+          await attachMatchingAssets(tx, {
```
buggyhuy, line 147: In attachMatchingAssets, A DeviceGroupMatching with productId: null covers every product from that manufacturer, so "all Siemens devices" is a normal plan input. 
- Maybe add a timeout for attachMatchingAssets?
- If we know there only a set number of maximum assets we can have under deviceGroup. We can also set a maximum assets so we done loop through all assets for the worst case.

src/features/tracking/server/routers.ts
```diff
@@ -991,15 +994,16 @@ export const trackingRouter = createTRPCRouter({
...
+   * per-asset child of a parent ticket that carries the proposal's
+   * toolCallId, same as any other asset attachment.
    */
   createFleetWorkOrder: protectedProcedure
     .input(
       z.object({
         /** The chat tool call this proposal came from — the idempotency key. */
         toolCallId: z.string(),
-        assetIds: z.array(z.string()).min(1),
+        assetIds: z.array(z.string()).min(1).max(50),
```
perrydev17: just checking: I recall seeing other places in the code base using the same validation for workOrder assetIds, do we need to apply the same?

perrydev17: like do we need the same here: https://github.com/PATCH-UPGRADE/viper/blob/main/src/features/chat/viper-agent/langgraph/tools.ts#L147

src/features/tracking/types.ts
```diff
@@ -335,7 +345,7 @@ export const workOrderListItemSchema = z.object({
   sourceLabel: z.string().nullable(),
   departments: z.array(departmentItemSchema),
   assignee: assigneeItemSchema.nullable(),
-  assets: z.array(linkedAssetSchema),
+  assets: z.array(assetTicketItemSchema),
```
buggyhuy, line 348: Just make sure nothing is breaking. I think thi changes the shape of the existing API.

prisma/schema.prisma
0xcad: Previously: One work order ticket could be linked to 5 individual assets. But how did we get granular progress reporting on those 5 individual assets?

Now: One work order ticket is liked to 5 AssetTicket models. Each AssetTicket bundles a work order and an asset together. You complete this ticket for one asset, mark that one work order as done. 

This is distinct from child tickets, which is still behavior we want to replicate from other platforms (an AssetTicket has just _one_ asset attached to it)

prisma/seed.ts
0xcad: @trunggo9000 I would take Perry's advice here, and

* don't import `serverOnlyStub`. define your helper functions in such a way that you do not have to import from a file marked server only
    * alternatively, drop `import "server-only"` from `asset-tickets.ts`

src/features/tracking/server/asset-tickets.ts
```diff
@@ -0,0 +1,171 @@
...
+    actorId,
+    assetId,
+    "attached",
+    asset,
+  );
+
+  return child.id;
+}
+
+export async function attachMatchingAssets(
```
0xcad, line 87: Add a comment above this function describing what it does for readability

e.g, 
// Given an array of deviceGroupMatching ID's, finds all assets. Create `AssetTicket`s with parent set to `parentTicketId`

src/features/mitigation/server/routers.ts
```diff
@@ -128,10 +133,57 @@ export const mitigationRouter = createTRPCRouter({
...
+              if (!matchingAppliesToDeviceGroup(matching, asset.deviceGroup))
+                continue;
+              await createAssetTicket(tx, {
+                parentTicketId: ticket.id,
+                assetId: asset.id,
+                actorId: ctx.auth.user.id,
+              });
+            }
+          }
+        }
```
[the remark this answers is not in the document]
0xcad: @trunggo9000 Please make the proposed fix

src/features/tracking/server/routers.ts
0xcad: @trunggo9000 I think what Perry is saying here is that if you add `.max(50)` on `createFleetWorkOrder` (the trpc endpoint), you should add it to the langgraph tool that creates fleet work orders in the first place

My personal take: I don't think you need the `.max(50)` at all, I would just remove it from here

If you did keep `.max(50)`, then I would take Perry's advice and move `.max(50)` to the tool call as well (what would probably be better entirely would be to make sure that tool call and the trpc endpoint used the same zod schema for assets, so we didn't have to have two copies)

0xcad (approved):
A few requested changes:
* Don't import `serverOnlyStub` in seed.ts
* Make the coderabbit proposed fix of deduplicating assetIds with a set
* I think you can remove `.max(50)` from the asset schema
* Add quick comments for documentation in asset-tickets.ts

PR #213, opened 2026-08-09
[VW-421] Integrate Question and Email agent with latest vendor/manufacturer/vendorcontact models (3/3)

perrydev17:
ticket:  https://northeastern-patch.atlassian.net/browse/VW-421

## Summary
- Escalation email suggestions now use device, manufacturer, vendor, contract, and contact information to recommend appropriate recipients.
- Added vendor and contact selection for generated escalation emails, with manufacturer fallback when vendor coverage is unavailable.
- Recipient editing now supports contact selection or manual comma-separated email entry.
- Refreshed suggested emails when a question is marked as unsure.
- Improved recipient filtering and handling of unavailable vendor coverage.

### Flow
email-flow

### Draft email To field pull correct sample vendors contacts,
https://github.com/user-attachments/assets/4f8382df-7664-4430-8d8f-163de99681ac

### Manually add a recipient

https://github.com/user-attachments/assets/c9101c30-0b01-49eb-85e2-288c4cbc1fcb

### DIFF
--- a/scripts/seed-notifications.ts
+++ b/scripts/seed-notifications.ts
@@ -13,6 +13,7 @@ import {
   VersionStatus,
 } from "@/generated/prisma";
 import prisma from "../src/lib/db";
+import { AsteriskSquare } from "lucide-react";
 
 const SEED_USER = {
   email: "user@example.com",
@@ -80,6 +81,19 @@ const SYNGO_PLAZA_ASSETS = [
     role: "PACS Workstation",
     networkSegment: "RADIOLOGY-PACS",
   },
+];
+
+// Mock differnt use cases
+const QUESTION_CASES = [
+  {
+    product: "syngo.via",
+    expect: "VENDOR GE Healthcare - dropdown, 2 contacts",
+  },
+  {
+    product: "MAGNETOM FAMILY",
+    expect: "VENDOR Philips - dropdown, 1 contact",
+  },
+  { product: "Symbia Intevo", expect: "MANUFACTURER - no contacts, input" },
 ];
 
 async function seedSyngoPlazaVexScenario(userId: string) {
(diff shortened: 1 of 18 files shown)


scripts/seed-notifications.ts
```diff
@@ -765,6 +778,247 @@ The application deserialises untrusted data without sufficient validations that
...
+async function coverProducts(
+  vendorId: string,
+  title: string,
+  products: string[],
+) {
+  const assets = await prisma.asset.findMany({
+    where: {
+      deviceGroup: {
+        product: {
+          canonicalDisplayName: {
```
buggyhuy: I think `canonicalName` is the lower case version. Where `canonicalDisplayName` is not lower case.

src/features/questions/server/routers.ts
```diff
@@ -161,20 +137,29 @@ export const questionsRouter = createTRPCRouter({
...
-            ...draft,
-            body: `${draft.body.trimEnd()}\n${signature}`,
-            contacts,
-            toEmails: primary ? [primary.email] : [],
-          };
-        }),
+      const drafted = await Promise.all(
+        questions.map(
+          async (question): Promise<SuggestedVendorEmail | null> => {
+            const context = await gatherEscalationContext(question);
```
buggyhuy: Just curious, what happened if these calls failed?

buggyhuy (changes requested):

src/features/questions/agent/escalationEmail/schema.ts
```diff
@@ -1,12 +1,43 @@
...
+  vendorIds: string[],
+  contactIds: string[],
+) {
+  const vendorId =
+    vendorIds.length > 0
+      ? z.enum(vendorIds as [string, ...string[]]).nullable()
+      : z.null();
+  const contactId =
+    contactIds.length > 0
+      ? z.enum(vendorIds as [string, ...string[]])
```
buggyhuy: I feel like this should have been
```suggestion
      ? z.enum(contactIds as [string, ...string[]])
```

buggyhuy (changes requested):

src/features/questions/agent/escalationEmail/schema.ts
perrydev17: you are right, thanks for catching it. fixed

scripts/seed-notifications.ts
perrydev17: thanks, updated

src/features/questions/server/routers.ts
perrydev17: if either gatherEscalationContext or draftEscalationEmail fails, Promise.all reject and procedure 500 -> Question tab will see the default blank screen `No questions for...`. I added a try catch block and return null for the catch, so it will see the same default blank screen.

buggyhuy: sounds good, Good to know how its handled in the frontend.

src/features/questions/agent/escalationEmail/context.ts
```diff
@@ -1,15 +1,147 @@
...
+
+1. vendorId - who can actually answer?
+- Pick a VENDOR id when the answer depends on what is actually deployed, installed, configured, patched, or scheduled on the hospital's own machines,
+which version is really running, when the fleet can be taken offline, whether a change was already applied during past service, who owns the maintenance window.
+- Use null to email the MANUFACTURER - the company that built the device, only when the answer is a property of the product itself and no one servicing the fleet
+could know it - whether a vulnerable component ships in the product at all, whether the flaw is reachable given the product's design, whether a patch or VEX statement exists. 
+- Tie-breaker: if a vendor is listed and the question could plausibly be answered by whoever services the fleet, choose VENDOR.
+- We hold contact address for vendors but not for manufacturers, so MANUFACTURER leaves the user to find an address themselves. Do not choose it by default.
+- A vendor may carry the same company name as the manufacturer. That does not make it the manufacturer - judge by the question, not the name.
+- if no vendor is listed, choose MANUFACTURER. You have no other option.
```
0xcad: > if no vendor is listed, choose MANUFACTURER. You have no other option

Again, I would consider making the system prompt the output of a function, that is aware if `vendors` exist or not. If there are not vendors, then telling the model about vendors, giving tie breaking rules, is not necessary and potentially confusing.

src/features/questions/agent/escalationEmail/context.ts
```diff
@@ -1,15 +1,147 @@
...
+                .map((contact) => {
+                  const title = contact.title ? `${contact.title}` : "";
+                  return ` - id: ${contact.id} | ${contact.name} | ${title} | ${contact.email}`;
+                })
+                .join("\n")
+            : " (none on file)";
+          return `-id: ${vendor.id} | ${vendor.displayName}\n contacts:\n${contacts}`;
+        })
+        .join("\n")
+    : "No vendor is under contract for these assets, so MANUFACTURER is the only option.";
```
0xcad: Suggestion: If `vendors` is empty, make sure the Zod schema that the model gets reflects that MANUFACTURER is the only option. 

I would also remove `VENDOR - under contract to service these assets` from the prompt if this is the  case. An easy way to do that is to just move this string to `vendorSection`, and have `vendorSection` be null

**This leads me to a larger question:**
Currently, I believe vendor / manufacturer is left solely up to `escalationEmail`. However, consider the case where we have no vendors, and the agent asks a question that could not reasonably be handled by a device manufacturer (e.g, is this specific user-configurable option turned on, or off, on our machines). 

If the device was managed by a vendor, that could plausibly be answered, but not by a manufacturer.

The escalation email agent gets forced drafting an email then for this audience.

Personally, I think it may be better (although a larger rewrite) to do the following:
* on each question, store an audience (vendor or manufacturer). So the question agent drafts a question, and that agent chooses the escalation audience. Doesn't have to choose a specific vendor at this point, just essentially whether or not the question can be answered by a manufacturer
* If a question is marked unsure and it cannot be sent to a manufacturer, *and* the hospital has no vendors -- then do not fire the escalation email agent. Instead, just keep the question in the UI and show that "unsure" was selected. This shows that the user has incomplete information, they can answer it later if they get more info, and it prevents an unnecessary email from being drafted

What do you think?

src/features/questions/agent/escalationEmail/context.ts
```diff
@@ -1,15 +1,147 @@
...
+lead handles scheduling and contractual questions. Choosing nobody is valid and often correct - return an empty array when no listed person clearly fits.
+
+3. reasonWhy, subject, body.
+
+WRITING:
+- reasonWhy is INTERNAL. It appears in the app under "Why send this:" and is never sent. One or two sentences to colleague on what answering this unblocks.
+Do not write it as part of the email.
+- subject names the product and the CVE.
+- body is plain text: address the company by exactly the name given, say who is writing and why, state the question plainly, and say what form of answer helps
+(a version confirmation, a VEX statement, an affected-model list). Under 200 words, direct and factual, no deadlines or pressure. End with a sign-off line such
```
0xcad: "VEX" is a cybersecurity industry term and may not be relevant or known to a vendor (some vendors do use VEX, most don't)

I actually gave Huy a ticket last sprint asking him to remove references to "VEX" to the `mitigate`, `triage`, and `VEX` agents:
* https://northeastern-patch.atlassian.net/browse/VW-390

src/features/questions/agent/escalationEmail/schema.ts
```diff
@@ -1,12 +1,43 @@
...
+      : z.null();
+  const contactId =
+    contactIds.length > 0
+      ? z.enum(contactIds as [string, ...string[]])
+      : z.string();
 
-export type EscalationDraft = z.infer<typeof escalationEmailschema>;
+  return z.object({
+    vendorId: vendorId.describe(
+      "Id of the chosen vendor to email. frp, tje VEMDPR list. Null means email the manufacturer instead, choose that only when no listed vendor could answer the question, or when no vendors are listed.",
```
0xcad: Coderabbit also flags the typo here

src/features/questions/agent/escalationEmail/schema.ts
```diff
@@ -1,12 +1,43 @@
...
+    ),
+    contactIds: z
+      .array(contactId)
+      .describe(
+        "Contact ids from the chosen vendor. Empty array when no listed contact is clearly fits, and always when vendorId is null.",
```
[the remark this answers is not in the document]
0xcad: Flagging this

0xcad (changes requested):
see comments

src/features/questions/agent/escalationEmail/context.ts
```diff
@@ -110,38 +117,26 @@ export async function gatherEscalationContext(
...
+3. Write reasonWhy, subject, body.`;
 
-WRITING:
-- reasonWhy is INTERNAL. It appears in the app under "Why send this:" and is never sent. One or two sentences to colleague on what answering this unblocks.
-Do not write it as part of the email.
-- subject names the product and the CVE.
-- body is plain text: address the company by exactly the name given, say who is writing and why, state the question plainly, and say what form of answer helps
-(a version confirmation, a VEX statement, an affected-model list). Under 200 words, direct and factual, no deadlines or pressure. End with a sign-off line such
-as "Best regards," and nothing after it. The app appends the sender's name, so a name you write will appear twice.
+const TO_MANUFACTURER = `This email goes to the manufacturer that built the device. Set vendorId to null and contactIds to an empty array - we hold no manufacturer addresses, so the user will supply one. Write reasonWhy, subject, body.`;
```
0xcad, line 129: > Set vendorId to null and contactIds to an empty array

I'm also looking at `src/features/questions/agent/escalationEmail/schema.ts`

In my preference, I would've again just removed the `vendorId` and `contactIds` fields _entirely_ from structured output, rather than set that to null. Does langchain let you do that?

I think the code could be a little more clean if we didn't force a model (or ask one, in the system prompt), to supply `null` / an empty array, when we know that we don't except those fields.

So I'd case on audience, and return two different zod objects. You can use zod's `.extend` for the shared reasonWhy, subject, and body fields 

Non-blocking though this is small

0xcad (approved):

buggyhuy (approved):

PR #221, opened 2026-08-12
[VW-428] Moved ai / partner integrations into new platform modules

0xcad:
* Created a `core` directory for integrations
* Converted `ai` and `partner` integrations into explicit platforms
* Difference from original RFC: No longer a shared `Session` type, each platform will handle its own session logic. The difference between basic HTTP header authentication and what we do for teamplay Fleet was quite substantial, and forcing new platforms to fit one session interface may be overly prescriptive 

**Code breakdown** (because +3,191 is a lot):

* Added +1,437 lines of code in tests (previously, our ai and partner integrations only tested `routers.ts` files) 
* +1,530 / -764 lines of production code -- a lot of moving code, then 800 LOC for scaffolding for other platforms
* +224 / −150 lines of documentation for LLM's to use in subsequent PR's

**Verification:**

* I was able to copy 326 assets from Blueflow to VIPER via the partner integration

image

* I modified the github VIPER<->Blueflow verification ci job to ignore if VIPER fails to typecheck, still build it, and run the integration. behold:

image

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


.github/workflows/blueflow-integration.yml
0xcad, line 1: blueflow integration ci job now ignores type errors to test mid-refactor

docs/integrations-redesign-implementation.md
0xcad, line 1: docs for an LLM to make notes as the refactor gets built out. I figured it would be helpful for giving context to other LLM's faster (or quick summary for humans)

docs/integrations-redesign.md
0xcad, line 1: I removed a Session generic type for platforms, as I don't think that can be cleanly re-used between platforms. instead, each platform will have to define its own helpers for auth, and reuse those between modules

src/features/assets/server/routers.ts
0xcad, line 1: all of these `src/features/assets|vulnerabilities|remediations|etc/server/routers.ts` files have the same changes. the token used to authenticate partner integrations now returns what resource it was scoped for.

also, we moved `upstreamApi` and `webUrl` to `External*Mapping`. However, Blueflow still gives us those fields per asset. So we have to remove those fields (we do this as

```
 upstreamApi: _upstreamApi,
 webUrl: _webUrl,
```
)

before doing anything with `itemData`

src/features/integrations/core/__tests__/callback.test.ts
0xcad, line 1: see callback.ts

src/features/integrations/core/__tests__/credentials.test.ts
0xcad, line 1: see credentials.ts

src/features/integrations/core/__tests__/registry.test.ts
0xcad, line 1: see registry.ts

src/features/integrations/core/__tests__/urls.test.ts
0xcad, line 1: see urls.ts

src/features/integrations/core/sync/cadence.ts
0xcad, line 1: How often a resource syncs, and how far to back off when it fails.

We store in each `IntegrationResourceSync` row a `nextSyncAt` datetime to schedule the next sync, this computes that

src/features/integrations/core/sync/resources.ts
0xcad, line 1: simple helpers to determine which resource modules are handled on an integration

src/features/integrations/core/sync/upsert.ts
0xcad, line 1: Moved a bunch of logic from `src/lib/router-utils.ts` to here

After we get data from a platform, how do we dedup/upsert it using External*Mapping tables

src/features/integrations/core/callback.ts
0xcad, line 1: used only by ai and partner integrations. see top comment

src/features/integrations/core/credentials.ts
0xcad, line 1: We store auth details encrypted in the Integration table as bytes

This is an improvement from just saving auth in the clear in our db. the test cases for this confirm that if there's no auth for an integration, credentials gets saved as `null`, and that we salt encrypted data

src/features/integrations/core/registry.ts
0xcad, line 1: exports a list of platforms in VIPER. teamplay fleet will be registered here in VW-431

src/features/integrations/core/types.ts
0xcad, line 1: defines common interfaces that platforms and resource modules inherit / compose from

src/features/integrations/core/types.ts
```diff
@@ -0,0 +1,128 @@
...
+// TODO: VW-431, use a cursor to store pagination logic when polling data from fleet?
+// if you use it, great, keep it, if not, just delete it
+export type Cursor = unknown;
+
+export interface Page<TRaw> {
+  items: TRaw[];
+  /** The cursor to resume from *after* this page. null = end of stream. */
+  cursor: Cursor | null;
+}
+
```
0xcad, line 23: For @perrydev17 and @markglose-bc in particular:

One idea was to use this `Cursor` type to handle pagination / any non-credential state saving. It's raw JSON that gets saved to the db. So for the work order integration, save any filters that need to be applied to the API endpoint (ordering, paging, specific timestamps for future syncs) to the db for durability.

If you use `Cursor` this way, great! If no one uses it or needs it, let's remove the type, and possibly have a db migration to remove it from prisma tables, too

src/features/integrations/core/types.ts
```diff
@@ -0,0 +1,128 @@
...
+  schema: Record<string, unknown>;
+}
+
+// TODO: VW-431 Fleet should use this on resource modules
+export interface UrlBuilders<TConfig = unknown> {
+  /** What's the API URL for this record on their platform? */
+  apiUrlFor?(externalId: string, config: TConfig): string | null;
+  /** Where does a human look to find this record? */
+  webUrlFor?(externalId: string, config: TConfig): string | null;
+}
```
0xcad, line 45: @markglose-bc @perrydev17 @buggyhuy 

A resource module implements this interface to convert externalId's into API/web URLs

Used for:
* getting a work order detail API link from an external ID
* getting a work order detail page/frontend page, from an external ID
* doing the same thing with assets/notifications

src/features/integrations/core/types.ts
```diff
@@ -0,0 +1,128 @@
...
+  // TODO: VW-431: can change `listChanged` schema if it doesn't work for fleet
+  get(externalId: string): Promise<TRaw>;
+  toCanonical(raw: TRaw, config: TConfig): TCanonical;
+
+  // we push to their platform
+  create?(draft: TCanonical): Promise<{ externalId: string; raw: TRaw }>;
+  update?(
+    externalId: string,
+    patch: Partial<TCanonical>,
+  ): Promise<{ externalId: string; raw: TRaw }>;
```
0xcad, line 64: @markglose-bc 

implement these functions to transform VIPER work orders into teamplay Fleet work orders

src/features/integrations/core/types.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  /** how often this resource should sync, in seconds. null = no opinion. */
+  defaultSyncEvery: number | null;
```
0xcad, line 67: e.g, teamplay fleet assets sync daily, work orders sync every 10 minutes. something like that, we can tune numbers later

src/features/integrations/core/types.ts
```diff
@@ -0,0 +1,128 @@
...
+export interface ConnectorDefinition<TConfig, TCreds> {
+  platform: PlatformEnum;
+  displayName: string;
+  /** Validates `Integration.config`. */
+  configSchema: z.ZodType<TConfig>;
+  /** Validates the decrypted `Integration.credentials`. */
+  credentialSchema: z.ZodType<TCreds>;
+}
+
+export interface ConnectorModule<TConfig = unknown, TCreds = unknown> {
```
0xcad, line 107: a Platform defines the ConnectorModule interface

src/features/integrations/platforms/ai/config.ts
```diff
@@ -0,0 +1,17 @@
...
+import { authCredentialSchema } from "@/features/integrations/core/credentials";
+import { genericConfigSchema } from "@/features/integrations/core/sync/resources";
+import { safeUrlSchema } from "@/lib/schemas";
+
+/**
+ * What an operator has to provide for an AI integration.
+ */
+export const configSchema = genericConfigSchema.extend({
+  integrationUri: safeUrlSchema,
+  additionalInstructions: z.string().optional(), // given to n8n agent
```
0xcad, line 11: was Integration.prompt

src/features/integrations/platforms/partner/sync.ts
0xcad, line 1: previously lived in inngest function, syncPartnerIntegration

src/features/integrations/platforms/ai/sync.ts
0xcad, line 1: previously lived in inngest function, syncAIIntegration

src/features/integrations/types.ts
```diff
@@ -1,82 +1,78 @@
...
+
+/**
+ * What `integrations.create` / `.update` accept.
+ *
+ * `config` stays opaque here on purpose. Narrowing it per-platform would mean
+ * importing every platform's `configSchema`, and those reach `core/credentials.ts`
+ * (`node:crypto`, `server-only`) — which would make this module unusable from a
+ * client component. The router narrows it instead, with the platform's own
+ * schema, so the platform stays the single validator either way.
+ */
```
0xcad, line 49: Thought for review: Should we restructure our imports so that we can use per-platform configSchema in the zod schema to validate integrations? Currently we run into a problem where this is used in client modules, so it can't import from a server only module

Personally I don't think it's necessary to move this logic here if it's enforced elsewhere on the backend, but that's an option, too

src/features/tracking/types.ts
0xcad, line 1: downstream of VW-427, getting types to pass with new schema changes

src/features/tracking/server/routers.ts
0xcad, line 1: downstream of VW-427, getting types to pass with new schema changes

src/features/vendors/server/routers.test.ts
0xcad, line 1: Contract now reaches assets through ManagesRelationship join table

src/inngest/functions/sync-integrations.ts
0xcad, line 1: moved a lot of the logic here into platform ai/sync.ts and partner/sync.ts

that should be way more scalable than having to define a new REST_MAPPER, or bespoke sync file, here every time -- one of the points of this refactor

src/lib/router-utils.ts
0xcad, line 1: moved most logic into core/upsert.ts

